Issue #2, opened 2021-12-11
Consolidate shim-proto from runD for kata v3.0


jiangliu:
There's an ongoing efforts to upstream Alibaba's runD for Kata 3.0. There's a similar implementation of shim-client/shim-proto from runD. It would be great to consolidate these two implementations.

AkihiroSuda:
What is runD?

fuweid:
> What is runD? 

runDragonball I think.

AkihiroSuda:
> > What is runD?
> 
> runDragonball I think.

So, what is runDragonball ? 😅
Is there any repo or doc?

fuweid:
> > > What is runD?
> 
> > 
> 
> > runDragonball I think.
> 
> 
> 
> So, what is runDragonball ? 😅
> 
> Is there any repo or doc?

https://medium.com/kata-containers/kata-containers-vptg-community-gathering-new-features-and-a-glimpse-of-kata-3-0-7ba025025e9e 

Section: Integrated Rust Runtime

Less docs are about this, since it is not open source. @AkihiroSuda

jiangliu:
> > > What is runD?
> > 
> > 
> > runDragonball I think.
> 
> So, what is runDragonball ? 😅 Is there any repo or doc?

Let me give some background.
Dragonball is a light-weight microvm written in rust, which is similar to firecracker/cloud hypervisor.
`runD` is a kata-runtime written in rust, which has integrate container runtime(shim-v2) and microvm together.
We are in the process to upstream runD to Kata Containers. Here are some references:
https://github.com/kata-containers/kata-containers/issues/3052
https://github.com/kata-containers/kata-containers/pull/3251/files

PR #3, opened 2021-12-11
Consolidate shim-proto from Alibaba runD into shim-client

jiangliu:
There's an ongoing efforts to upstream Alibaba's runD for Kata 3.0. There's a similar implementation of shim-client/shim-proto from runD. It would be great to consolidate these two implementations.

Fixes: #2


crates/shim-client/README.md
```diff
@@ -7,11 +7,32 @@
...
+
+```toml
+[dependencies]
+containerd-shim-client = "0.1"
+```
 
 Basic client code looks as follows:
 
 ```rust
+extern crate containerd_shim_client as client;
```
mxpv, line 34: nit: I think `extern` is no longer needed since these crates target 2018 edition: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-extern-crate

crates/shim-client/README.md
```diff
@@ -27,8 +48,11 @@ let resp = task_client.connect(context, &req)?;
 
 ## Example
 
-Have a look on example [here](./examples/connect.rs).
+- [ttRPC shim client](./examples/ttrpc-client.rs)
+- [ttRPC shim server](./examples/ttrpc-server.rs)
+- [ttRPC client connect](./examples/connect.rs).
```
mxpv, line 53: nit: TTRPC similarly to GRPC?

mxpv (approved):
A couple of minor nits, overall LGTM

crates/shim-client/README.md
jiangliu: Fixed, still think about rust 2015 sometimes:(

crates/shim-client/README.md
jiangliu: Pardon, didn't get your point here. Should I mention that GRPC is similar to ttRPC so the example may work for GRPC too?

mxpv: No, I meant like rename `ttRPC` to just `TTRPC`

mxpv: I'll go ahead and merge this one. It's not a blocker and easy to follow up.

PR #5, opened 2021-12-12
[RFC] Bugfix and enhancements for crate containerd-shim 

jiangliu:
This is an in-progress patch set for the containerd-shim crate, which 
1) fixes some resource leakage issues related to RawFd
2) adds more unit test cases
3) extends spawn() 
4) refine the Shim trait interfaces

### DIFF
--- a/crates/shim/src/lib.rs
+++ b/crates/shim/src/lib.rs
@@ -22,8 +22,10 @@ use std::error;
 use std::fs;
 use std::hash::Hasher;
 use std::io::{self, Write};
+use std::os::unix::io::AsRawFd;
 use std::os::unix::io::RawFd;
-use std::path::PathBuf;
+use std::os::unix::net::UnixListener;
+use std::path::{Path, PathBuf};
 use std::process::{self, Command, Stdio};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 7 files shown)


mxpv:
Could you please rebase against latest `main` to exclude commits that are already there from this PR.

jiangliu:
> Could you please rebase against latest `main` to exclude commits that are already there from this PR.

Rebased onto the latest main branch.

crates/shim/src/lib.rs
```diff
@@ -84,45 +87,52 @@ pub struct StartOpts {
...
 
-    fn new(
-        id: &str,
-        namespace: &str,
-        publisher: RemotePublisher,
-        config: &mut Config,
-        exit: ExitSignal,
-    ) -> Self;
+    /// Type to provide task service for the shim.
+    type T: Task + Send + Sync;
```
mxpv, line 129: I think we should move away from the concept that there 1:1 relation between shim and task service.
We started adding plugin support to shims, so any number of arbitrary service can now be registered (https://github.com/containerd/containerd/pull/5817, https://github.com/containerd/containerd/pull/6021).

jiangliu: Exactly, the same reason to refactor Shim interfaces.

crates/shim/src/lib.rs
```diff
@@ -84,45 +87,52 @@ pub struct StartOpts {
 }
 
 /// Helper structure that wraps atomic bool to signal shim server when to shutdown the TTRPC server.
+///
 /// Shim implementations are responsible for calling [`Self::signal`].
 #[derive(Clone)]
-pub struct ExitSignal(Arc<AtomicBool>);
+pub struct ExitSignal(Arc<(Mutex<bool>, Condvar)>);
```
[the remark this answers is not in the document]
jiangliu: The issue is on the wait side:
```
    /// Wait for the exit signal to be set.
    fn wait(&self) {
        while !self.0.load(Ordering::Acquire) {
            std::hint::spin_loop();
        }
    }

```

The wait() is effectively a busy loop, so it may waste huge CPU cycles.

PR #7, opened 2021-12-13
Update shim/README.md for running examples locally

liubin:
The README.md of shim crate is out-of-code, so update it.

And also fix a bug(the first commit) of passing the wrong container id to shim.

Now it depends on https://github.com/containerd/ttrpc-rust/pull/116 to let shim-client to use normal Unix domain socket.


jiangliu (approved):

crates/shim/README.md
```diff
@@ -66,26 +66,28 @@ impl shim::Task for Service {
 }
 
 fn main() {
-    shim::run::<Service>("io.containerd.empty.v1")
+    shim::run::<Service>("io.containerd.empty.v2")
```
fuweid, line 69: `io.containerd.${shim-name}.vX` is ok. Both v1 and v2 are valid version of shim.
I think we should not confuse user here.

liubin: I remember that there is a rule to map a shim name(`io.containerd.empty.v1`) to its binary(`containerd-shim-empty-v2`), the steps to run an empty that use the name with `containerd-shim-empty-v2`, so I think here should change to the same v2.

jiangliu: Seems the containerd-shim crate targets shim-v2 only?

fuweid: The version is just version of shim implementation, not for the shim-v2 proto. containerd doesn't map v1 to v2 so that the `io.containerd.empty.v1` will be `containerd-shim-empty-v1`. The related code is [here](https://github.com/containerd/containerd/blob/7fa44fc98f4bde30fa154c361e11609f742aaccb/runtime/v2/shim/util.go#L92)

crates/shim/README.md
```diff
@@ -66,26 +66,28 @@ impl shim::Task for Service {
...
 
+**Note**: All operations are in the root directory of `rust-extensions`.
+
 With shim v2 runtime:
 
 ```bash
-$ cargo build --example empty-shim
-$ sudo cp ./target/debug/examples/empty-shim /usr/local/bin/containerd-shim-empty-v2
+$ cargo build --example empty_shim
+$ sudo cp ./target/debug/examples/empty_shim /usr/local/bin/containerd-shim-empty-v2
```
fuweid, line 82: /usr/local/bin/containerd-shim-empty-v2 -> /usr/local/bin/containerd-shim-empty-v1?

liubin: @fuweid I update the sample code because this README uses `containerd-shim-empty-v2` and `io.containerd.empty.v2` in these commands, do you mean that if w keep them consistent, it doesn’t matter whether v1 or v2?

fuweid: yes. it doesn't matter whether v1 or v2. But I would like to keep it with v1 because empty is just released. 
It's just weird that the version starts from 2 not 1....

liubin: @fuweid I have updated as your suggested that change all `v2` to `v1`

mxpv (approved):

PR #9, opened 2021-12-14
shim: distinguish between runtime id and container id

jiangliu:
There's a misuse of runtime id and container id in Shim::new(). Both
runtime id and container id are needed by shim implentions, so add
parameter `container id` to Shim::new().

Fixes: #13

### DIFF
--- a/crates/shim/src/lib.rs
+++ b/crates/shim/src/lib.rs
@@ -129,7 +129,20 @@ pub trait Shim {
     type T: Task + Send + Sync;
 
     /// Create a new instance of Shim.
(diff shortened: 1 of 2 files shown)


crates/shim/src/lib.rs
```diff
@@ -129,7 +129,13 @@ pub trait Shim {
     type T: Task + Send + Sync;
 
     /// Create a new instance of Shim.
-    fn new(id: &str, namespace: &str, publisher: RemotePublisher, config: &mut Config) -> Self;
+    fn new(
+        runtime_id: &str,
+        id: &str,
```
mxpv, line 134: Could you please update docs for this function to clarify the difference between `runtime_id` and `id`?

mxpv (approved):

PR #11, opened 2021-12-17
Add tasks

liubin:
This PR includes two commits:

- add Task service
- add an example to create containers and tasks

### DIFF
--- a/crates/client/examples/container.rs
+++ b/crates/client/examples/container.rs
@@ -1,0 +1,165 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


crates/client/examples/container.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    // create temp dir for stdin/stdout/stderr
+    let tmp = std::env::temp_dir().join("containerd-client-test");
+    fs::create_dir_all(&tmp).expect("Failed to create temp directory");
+    let stdin = tmp.join("stdin");
+    let stdout = tmp.join("stdout");
+    let stderr = tmp.join("stderr");
+    let _ = File::create(&stdin).expect("Failed to create stdin");
+    let _ = File::create(&stdout).expect("Failed to create stdout");
+    let _ = File::create(&stderr).expect("Failed to create stderr");
```
mxpv, line 102: nit: no need in `let _ =` since we use `expect` ?

mxpv (commented):
Thanks! Overall LGTM, a couple of minor nits.

liubin:
@mxpv I have updated two places. 

And I also moved `with_namespace` to `src/lib.rs`, I think it is worth it.

PR #12, opened 2021-12-17
Very initial remote snapshotter crate

mxpv:
This PR adds:
- GRPC generated code.
- Snapshotter types.
- Server wrapper.
- Snapshotter example.
- README


jiangliu:
Interesting, maybe we could switch the nydus-snapshotter to rust in future:)
https://github.com/containerd/project/issues/83

PR #19, opened 2022-01-22
[shim] Set the "child subreaper" attribute

Burning1020:
The shim process can have "child subreaper" attribute when arg2 is **nonzero**.
see: https://man7.org/linux/man-pages/man2/prctl.2.html


fuweid (approved):
LGTM

mxpv (approved):

PR #20, opened 2022-01-22
[shim] Fix socket binding address

Burning1020:
As there is no need in "-socket" flag, we'd better use SOCKET_FD as the binding address of task server only.

If "-socket" flag was passed as socket path by the first start command, the second default command will failed with "Address in use" error because the socket was already passed to it by Fd mapping when spawn command. 

Signed-off-by: Burning <burning9699@gmail.com>

### DIFF
--- a/crates/shim/src/lib.rs
+++ b/crates/shim/src/lib.rs
@@ -253,11 +253,7 @@ where
             let task_service = create_task(Arc::new(Box::new(task)));
             let mut server = Server::new().register_service(task_service);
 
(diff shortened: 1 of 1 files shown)


mxpv:
Should we remove `socket` command line entirely?
@jiangliu @liubin

fuweid (approved):
LGTM

I think it is good start to remove socket option since the socket address has been manged by shim author.

fuweid:
ping @jiangliu and @liubin

PR #22, opened 2022-02-03
runc client

Yuna-Tomi:
runc client 
Signed-off-by: Yuna Tomida <ytomida.mmm@gmail.com>


Yuna-Tomi:
- runc client crate, corresponds to [go-runc](https://github.com/containerd/go-runc) in Go
- Thanks a lot to earlier efforts [rust-runc](https://github.com/pwFoo/rust-runc)
    - It's archived then I re-implement based on it.
- Contains runc client of sync/async version
- Supported
    - runc create / start / kill / delete / state
    - exec is available only in sync client now
- What left unimplemented
    - checkpoint related commands
    - console utilities (ref: [Go version](https://github.com/containerd/go-runc/blob/main/console.go))

Yuna-Tomi:
And updated MSRV to 1.54 for using `oci-spec` crate: https://github.com/containers/oci-spec-rs

AkihiroSuda:
LGTM, but needs to fix the format error
```
Run cargo fmt --all -- --check --files-with-diff
/home/runner/work/rust-extensions/rust-extensions/crates/runc-rust/src/lib.rs
Error: Process completed with exit code 1.
```
https://github.com/containerd/rust-extensions/runs/5048032832?check_suite_focus=true

Thanks!

Yuna-Tomi:
I fixed it on local, but I want to ask something before push it.
When I run `clippy` with options set on CI, this error came out:
スクリーンショット 2022-02-03 20 40 17
Actually it's not what I implemented here, but is it fine fixing on this PR?

AkihiroSuda:
> Actually it's not what I implemented here, but is it fine fixing on this PR?
Maybe I can fix renaming it to ParseFailed.

Yes, in a separate commit 🙏 Thanks

mxpv:
> Actually it's not what I implemented here, but is it fine fixing on this PR?
Maybe I can fix renaming it to ParseFailed.

I think this was introduced after updating rust to 1.54 in [rust-toolchain.toml](https://github.com/containerd/rust-extensions/pull/22/files#diff-2b1bde2cf3a858b7bf7424cb8bcbf01f35b94dc80b925d9432cbab3319ca9b4eR2)

Cargo.toml
```diff
@@ -4,5 +4,6 @@ members = [
     "crates/logging",
     "crates/shim-protos",
     "crates/shim",
-    "crates/snapshots"
+    "crates/snapshots",
+    "crates/runc-rust"
```
mxpv, line 8: just `crates/runc` ?

kzys: +1

crates/runc-rust/src/container.rs
```diff
@@ -0,0 +1,77 @@
...
+                \"bundle\": \"/path/to/root\",
+                \"rootfs\": \"/path/to/rootfs\",
+                \"created\": 1431684000,
+                \"annotations\": {
+                    \"foo\": \"bar\"
+                }
+            }";
+
+        let c: Container = serde_json::from_str(j).unwrap();
+        println!("{:#?}", c);
```
kzys, line 75: Personally prefer to keep stdout clean. If you have some properties you want to check, write them as tests.

crates/runc-rust/src/io.rs
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    // tokio version of set()
+    fn set_tk(&self, _cmd: &mut tokio::process::Command) -> std::io::Result<()> {
+        unimplemented!()
+    }
+
+    fn close_after_start(&self) {
+        unimplemented!()
+    }
```
kzys, line 60: Not clear about these `unimplemented`. Due to time constraints?

crates/runc-rust/src/io.rs
```diff
@@ -0,0 +1,333 @@
...
+}
+
+#[derive(Debug)]
+pub struct RuncPipedIO {
+    stdin: Option<Pipe>,
+    stdout: Option<Pipe>,
+    stderr: Option<Pipe>,
+}
+
+impl RuncPipedIO {
```
mxpv, line 141: I think we could omit these `Runc` prefixes everywhere.
`runc::RuncPipedIO` -> `runc::PipedIo`
Same for other structs.

crates/runc-rust/src/io.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    pub fn close_read(&self) {
+        let mut m = self.rd.lock().unwrap();
+        let _ = m.take();
+    }
+
+    pub fn close_write(&self) {
+        let mut m = self.wr.lock().unwrap();
+        let _ = m.take();
+    }
```
kzys, line 126: lock() returns Result. Does it make sense to return the errors rather than panic?

crates/runc-rust/src/io.rs
```diff
@@ -0,0 +1,333 @@
...
+
+#[derive(Debug)]
+pub struct RuncPipedIO {
+    stdin: Option<Pipe>,
+    stdout: Option<Pipe>,
+    stderr: Option<Pipe>,
+}
+
+impl RuncPipedIO {
+    pub fn new(uid: isize, gid: isize, opts: IOOption) -> std::io::Result<Self> {
```
kzys, line 142: Should this use isize here?

crates/runc-rust/src/lib.rs
```diff
@@ -0,0 +1,1019 @@
...
+    pub fn update(&self, id: &str, resources: &Linux) -> Result<()> {
+        let (mut temp_file, file_name): (NamedTempFile, String) =
+            utils::make_temp_file_in_runtime_dir()?;
+        {
+            let f = temp_file.as_file_mut();
+            let spec_json =
+                serde_json::to_string(resources).map_err(Error::JsonDeserializationFailed)?;
+            f.write(spec_json.as_bytes())
+                .map_err(Error::SpecFileCreationError)?;
+            f.flush().map_err(Error::SpecFileCreationError)?;
```
mxpv, line 427: This block can be reduced to:
```rust
fs::write(path, spec_json)?;
```

kzys (changes requested):

crates/runc-rust/src/io.rs
mxpv: not in case of std `Mutex` - https://users.rust-lang.org/t/should-i-unwrap-a-mutex-lock/61519

crates/runc-rust/src/container.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub annotations: HashMap<String, String>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn serde_test() {
+        let j = "
```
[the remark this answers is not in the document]
Yuna-Tomi: It's good, thank you!

Cargo.toml
Yuna-Tomi: Ok, thank you!

crates/runc-rust/src/io.rs
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+}
+
+/// This struct represents pipe that can be used to transfer
+/// stdio inputs and outputs
+/// when one side of closed, this struct represent it with [`None`]
+#[derive(Debug)]
+pub struct Pipe {
+    // Might be ugly hack: using mutex in order to take rd/wr under immutable [`Pipe`]
```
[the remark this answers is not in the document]
Yuna-Tomi: Actually what I want to do is accessing struct containing `Pipe` like `PipeIO` here from multipile thread, and extract file, and copy the stdio to other io entry from higher layer(e.g. Fifo that passed from containerd to shim).
In such case, we should `Arc` it to reference from multiple thread and need `Mutex` to be mutable(to read/write), but any two threads will not access the same entry(e.g. stdout and stdout) and I thought it's redundant to make `Mutex<PipedIO>`.
Is there any other good way to do this?

crates/runc-rust/src/io.rs
```diff
@@ -0,0 +1,333 @@
...
+    fn stdout(&self) -> Option<File> {
+        None
+    }
+
+    /// Return read side of stderr
+    fn stderr(&self) -> Option<File> {
+        None
+    }
+
+    fn close(&self) {
```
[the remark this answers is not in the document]
Yuna-Tomi: Ok, thank you!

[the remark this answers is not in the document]
Yuna-Tomi: As destructor automatically drop files inside structs, I will just erase this trait method.

crates/runc-rust/src/container.rs
Yuna-Tomi: Thank you for recommend but I'm sorry, I couldn't see your point.
This function is in `mod tests` and attributed as `#[test]`, but it's not enough?

crates/runc-rust/README.md
```diff
@@ -0,0 +1,43 @@
...
+This crate is based on archived [rust-runc](https://github.com/pwFoo/rust-runc).
+
+## Usage
+Both sync/async version is available.
+You can build runc client with `RuncConfig` in method chaining style.
+Call `build()` or `build_async()` to get client.
+Note that async client depends on [tokio](https://github.com/tokio-rs/tokio), then please use it on tokio runtime.
+
+```rust
+use runc_rust as runc;
```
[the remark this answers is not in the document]
Yuna-Tomi: Ok, thank you!

crates/runc-rust/src/io.rs
Yuna-Tomi: Ok, thank you!

crates/runc-rust/src/lib.rs
Yuna-Tomi: Ok, thank you!

crates/runc-rust/src/io.rs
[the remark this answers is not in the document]
Yuna-Tomi: Also, like `close_after_write`, I want to close some files(but not all) manually from threads.

crates/runc-rust/src/container.rs
kzys: What you'd like to do with the `println!` output? If you want to sometimes manually check the result, it would be better to put more context or use `dbg!`.

That being said, it would be great if you can automate the manual checking part by writing assertions.

crates/runc-rust/src/io.rs
```diff
@@ -0,0 +1,333 @@
...
+        unimplemented!()
+    }
+
+    // tokio version of set()
+    fn set_tk(&self, _cmd: &mut tokio::process::Command) -> std::io::Result<()> {
+        unimplemented!()
+    }
+
+    fn close_after_start(&self) {
+        unimplemented!()
```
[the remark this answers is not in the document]
Yuna-Tomi: Ok, thank you!

crates/runc-rust/src/io.rs
Yuna-Tomi: I' ll remove them, thank you!

crates/runc-rust/src/io.rs
Yuna-Tomi: Actually, I just followed [Go's](https://github.com/containerd/go-runc/blob/main/io.go#L83) but it seems to be unnecessary...

Yuna-Tomi: Thank you, I'll change it to `u32`

crates/runc-rust/src/container.rs
Yuna-Tomi: Oh, I see, thank you!
I'll rewrite them with assertions.

AkihiroSuda (approved):
CI green.

Let's merge this toward reopening the actual shim part (https://github.com/containerd/rust-extensions/pull/21)

mxpv:
@AkihiroSuda do we need 6e17be8d75f72f6f83c4430763a26b1af5afe0f5 for this PR? Seems like runc has no dependencies on these.

mxpv (approved):

mxpv:
@kzys PTAL

AkihiroSuda:
> @AkihiroSuda do we need [6e17be8](https://github.com/containerd/rust-extensions/commit/6e17be8d75f72f6f83c4430763a26b1af5afe0f5) for this PR? Seems like `runc` crate has no dependencies on these.

Cargo automatically bumped up them, so I committed.

.github/workflows/ci.yml
```diff
@@ -16,7 +16,12 @@ jobs:
       - run: cargo check --examples --tests --all-targets
       - run: cargo fmt --all -- --check --files-with-diff
       - run: cargo clippy --all-targets --all-features -- -D warnings
-      - run: cargo test --all-features
+      # runc::tests::test_exec needs $XDG_RUNTIME_DIR to be set
```
kzys, line 19: Shouldn't the test set the environment variable if so?

crates/runc/src/container.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use serde::{Deserialize, Serialize};
+use time::serde::timestamp;
+use time::OffsetDateTime;
+
+/// Information for runc container
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Container {
+    pub id: String,
+    pub pid: usize,
```
kzys, line 46: u32? Response's pid is u32.

crates/runc/src/lib.rs
```diff
@@ -0,0 +1,980 @@
...
+        Err(Error::Unimplemented("checkpoint".to_string()))
+    }
+
+    fn launch(&self, mut cmd: std::process::Command, combined_output: bool) -> Result<Response> {
+        let child = cmd.spawn().map_err(Error::ProcessSpawnFailed)?;
+        let pid = child.id();
+        let result = child.wait_with_output().map_err(Error::InvalidCommand)?;
+        let status = result.status;
+        let stdout = String::from_utf8(result.stdout).unwrap();
+        let stderr = String::from_utf8(result.stderr).unwrap();
```
kzys, line 237: Wouldn't it fail if the command outputs some non UTF-8 bytes (e.g. gzip)?

crates/runc/src/lib.rs
```diff
@@ -0,0 +1,980 @@
...
+        let _ = self.launch(self.command(&args)?, true).await?;
+        Ok(())
+    }
+
+    /// List all containers associated with this runc instance
+    pub async fn list(&self) -> Result<Vec<Container>> {
+        let args = ["list".to_string(), "--format-json".to_string()];
+        let res = self.launch(self.command(&args)?, true).await?;
+        let output = res.output.trim();
+        // Ugly hack to work around golang
```
kzys, line 568: I want to know more about this :)

crates/runc/src/lib.rs
```diff
@@ -0,0 +1,980 @@
...
+    fn fail_async_client() -> AsyncClient {
+        Config::new()
+            .command(CMD_FALSE)
+            .build_async()
+            .expect("unable to create runc instance")
+    }
+
+    fn dummy_process() -> Process {
+        serde_json::from_str(
+            "
```
kzys, line 709: raw string?

mxpv:
> Cargo automatically bumped up them, so I committed.

@AkihiroSuda can we revert it? Seems like its not relevant for this PR (also looks like it depends on environment, I'll TAL later).

kzys (commented):
Mostly nits. I'm fine merging the PR as is and address them later.

.github/workflows/ci.yml
AkihiroSuda: No, usually set by pam_systemd (which is missing in GHA)

AkihiroSuda:
> Seems like its not relevant for this PR (also looks like it depends on environment, I'll TAL later).

Not depending on local env, simply 2.27.1 is the latest ( https://crates.io/crates/protobuf ), and Cargo picks up the latest one because we do not commit Cargo.lock in the repo (why?)

mxpv:
> we do not commit Cargo.lock in the repo (why?)

We generally don't commit Cargo.lock for libraries: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

crates/runc/src/lib.rs
Yuna-Tomi: Ah, this is for just handling `null` json. But I think this way (with comparering output and "null") is not concise...
Is it comprehensive if I change it like below?
```rust
Ok(serde_json::from_str::<Option<Vec<usize>>>(output).map_err(/*omit*/)?.unwrap_or_default())
```

crates/runc/src/lib.rs
Yuna-Tomi: Sure. Thank you!

crates/runc/src/lib.rs
Yuna-Tomi: Sure. However which is preferable in such case, returning parse error or return the raw value to caller?
(This is just my personal opinion but I think returning error is better. Higher level wrappers which use this crate, like shim, will bend the stdout/err of command to IO that they prepare in advance. And they can handle such non-utf8 within their procedure)

crates/runc/src/container.rs
Yuna-Tomi: That's right, thank you!

Burning1020 (changes requested):

mxpv:
Ok, I'm going to merge this in. Let's address remaining items in follow up PRs.

PR #24, opened 2022-02-11
Minimum runc shim implementation

Burning1020:
Implement shim cli commands and some basic task APIs to run a runc container

Co-authored-by: Feng Shaobao <fshb1988@gmail.com>
Co-authored-by: Zhang Tianyang <burning9699@gmail.com>
Co-authored-by: Yu Qitao <yuqitao1024@qq.com>


Burning1020:
Hi, all! I'm from *Huawei Cloud Company*. My team has implemented a **whole** runc shim that can truly run a runc container and it works well with **containerd** and even **kubernetes** for mounths in production . 

Now that having noticed this "rust-extensions" project, we'd like to open our work and provide LTS as well. Actually, this commit is only a part just including some basic tasks(eg. *create/start/exec/wait/kill/delete/shutdown*...), the other tasks will be brought up if this commit can be accepted. Hope you can take our implementation into consinderations.

You can build it and replace to where the containerd-shim-runc-v2 (go verison) is for your own, use `ctr` or `crictl` to run pod/containers then I'm sure you'll get a wonderful experience!

Burning1020:
This commit has many features as you see, no limited to this:

1. The completed shim commands(*start/delete/default*) imps based on your current work
2. Impls for basic tasks, with runc cli underlay.
3. A signal handler wording with a static subscribe/nortify monitor and a subreaper
4. The abstract of ContainerFactory, Container, Process, CommandRuntime and related RuncFactory, RuncContainer, InitProcess/ExitProcess, Runc and so on. (we'll continuous optimize this part to make it more clear)
5. A checker for process exits and so on

Honestly, this commit may still have defects or something like that:

1. We've defined a **crate level error** in [error.rs](https://github.com/Burning1020/rust-extensions/blob/mini-runc/crates/shim/src/error.rs), which is different from your mod level error. The reason is that, it is much complicated to return each mod error layer by layer and there would left many duplicated errors. BTW, our solation is not perfect either, we can discuss this point lately.
2. Exec process with **tty** works well in rust 1.48 but in 1.52, the terminal display may be a little unfriendly but that doesn't affect function. I believe there are something wrong in `std::io::copy()`, I will figure out and push again at the end.
3. Not enouth UT and comments in our code.

There is still some work needed to be done, perhaps we can develop together!

Burning1020:
I'll fix the CI checks and continue to add changes lately, don't worry~

AkihiroSuda:
Seems a duplicate of https://github.com/containerd/rust-extensions/pull/21 and https://github.com/containerd/rust-extensions/pull/22

fuweid:
@AkihiroSuda I don't think it is duplicate 😄 I Checkout this patch and it provides more helpers like go-runc-shim, for instance，fifo scheme support and pure runC commandline. I would like to say it can work with #21 or #27 together. WDYT?

Cc @mxpv

Burning1020:
I just pushed again and it passed CI finally. 
cc plz. @AkihiroSuda @mxpv 
@jiangliu plz give review it to avoid same impls within we two

crates/runc/Cargo.toml
```diff
@@ -26,6 +26,7 @@ tempfile = "3.3.0"
 thiserror = "1.0.30"
 time = { version = "0.3.7", features = ["serde", "std"] }
 uuid = { version = "0.8.2", features = ["v4"] }
+containerd-shim = { path = "../shim", version = "0.2.0" }
```
mxpv, line 29: `runc` crate expected to be an independent from containerd CLI wrapper for runc binary. We should not depend on containerd-shim crate here.

crates/shim/src/io.rs
```diff
@@ -0,0 +1,289 @@
...
+use std::thread::JoinHandle;
+
+use crossbeam::sync::WaitGroup;
+use nix::sys::termios::Termios;
+
+use crate::error::{Error, Result};
+use crate::util::IntoOption;
+
+pub trait WriteCloser: Write {
+    fn close(&self);
```
mxpv, line 34: We generally should prefer `Drop` over `close`.

crates/runc/src/lib.rs
```diff
@@ -375,13 +371,12 @@ impl Runc {
             args.append(&mut opts.args()?);
         }
         args.push(id.to_string());
-        let mut cmd = self.command(&args)?;
+        let cmd = self.command(&args)?;
         match opts {
             Some(CreateOpts { io: Some(_io), .. }) => {
-                _io.set(&mut cmd).map_err(Error::UnavailableIO)?;
+                // _io.set(&mut cmd).map_err(Error::UnavailableIO)?;
```
mxpv, line 377: Can you pls add some clarity here?

crates/shim/src/lib.rs
```diff
@@ -180,29 +198,38 @@ pub trait Shim {
 }
 
 /// Shim entry point that must be invoked from `main`.
-pub fn run<T>(runtime_id: &str)
+pub fn run<T>(runtime_id: &str, opts: Vec<fn(&mut Config)>)
```
mxpv, line 201: Let's just pass `opts: Option<Config>` here
and below `opts.unwrap_or_default()`

Burning1020:
When handling container IO, we take a method that pass the stdio from request task to runc directly, which is quite different from pipeIO. Cause we think our solution has higher efficiency than CopyIO in shim layer.

crates/runc-shim/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+name = "containerd-shim-runc-v2"
+version = "0.1.0"
+authors = ["Shaobao Feng <fshb1988@gmail.com>", "Tianyang Zhang <burning9699@gmail.com>"]
+edition = "2018"
+license = "Apache-2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+async = []
```
mxpv, line 11: Async is not currently implemented in this PR, right? Should remove this flag?

crates/runc-shim/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#[macro_use]
+extern crate containerd_shim;
```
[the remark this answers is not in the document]
Burning1020: Great! older Rust in my team, i'll check it out in next PR.

crates/shim/src/io.rs
```diff
@@ -0,0 +1,299 @@
...
+impl IO for FIFO {
+    fn stdin(&self) -> Option<Box<dyn WriteCloser + Send + Sync>> {
+        None
+    }
+
+    fn stdout(&self) -> Option<Box<dyn Read + Send>> {
+        None
+    }
+
+    fn stderr(&self) -> Option<Box<dyn Read + Send>> {
```
mxpv, line 94: Should we just return `None` from the `Io` trait as default implementaiton and here just implement whats needs to be implemented?

crates/shim/src/io.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn set(&self, cmd: &mut Command) -> Result<()> {
+        let r: Option<std::io::Result<()>> = self.stdin.as_ref().map(|path| {
+            let stdin = OpenOptions::new()
+                .read(true)
+                .custom_flags(libc::O_NONBLOCK)
+                .open(path)?;
+            cmd.stdin(stdin);
+            Ok(())
+        });
+        r.unwrap_or(Ok(())).map_err(io_error!(e, "open stdin"))?;
```
mxpv, line 107:
```rust
        if let Some(ref path) = self.stdin {
            let stdin = OpenOptions::new()
                .read(true)
                .custom_flags(libc::O_NONBLOCK)
                .open(path)
                .map_err(io_error!(e, "open stdin"))?;
            cmd.stdin(stdin);
        }
```

crates/shim/src/monitor.rs
```diff
@@ -0,0 +1,182 @@
...
+pub(crate) struct Subscriber {
+    pub(crate) topic: Topic,
+    pub(crate) tx: Sender<ExitEvent>,
+}
+
+#[derive(Clone, Eq, Hash, PartialEq)]
+pub enum Topic {
+    PID(i32),
+    EXEC(String, String),
+    ALL,
```
mxpv, line 66: Pid, Exec, All

crates/shim/src/monitor.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum Subject {
+    // process pid
+    PID(i32),
+    // exec with containerd id and exec id for vm container,
+    // if exec is empty, then the event is for the container
+    EXEC(String, String),
```
mxpv, line 104: Pid, Exec

crates/runc-shim/src/main.rs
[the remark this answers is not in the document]
mxpv: Which version do you target? Rust 2015 seems to be quite old nowadays.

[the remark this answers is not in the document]
Burning1020: Never mind, it's my problem, I will remove it

crates/runc-shim/Cargo.toml
Burning1020: Just for reserved, remove it also OK

crates/runc/Cargo.toml
abel-von: maybe  we can change the runc crate to an "oci" crate, it's just a wrapper of runc binary call now, and runc is a standard implement of OCI runtime spec. we can change the command path from hard coded "runc" to an abstract method, say "get_command_path()", other oci compatible runtimes can easily use this crate to make their own shim-v2 binary, by implementing get_command_path()

Burning1020: > `runc` crate expected to be an independent from containerd CLI wrapper for runc binary. We should not depend on containerd-shim crate here.

`runc` crate needs IO and we take IO as a common trait for various runtime in `shim` crate, that's why it imports shim. I can follow your mind to put IO in runc as you always do. 
Could you wait for my new submit around this Fri possiably, otherwise I cam push in the followup PR.

mxpv: Sure, no rush, can wait till Friday.

crates/shim/src/lib.rs
Burning1020: `Option<Config>`? A single config or `None`, rather than `fn(&mut Config)` vectors:
https://github.com/containerd/containerd/blob/main/runtime/v2/shim/shim.go#L181

crates/runc-shim/src/runc.rs
```diff
@@ -0,0 +1,650 @@
...
+        Ok(container)
+    }
+}
+
+pub(crate) struct RuncContainer {
+    pub(crate) common: CommonContainer<InitProcess, ExecProcess>,
+}
+
+impl Container for RuncContainer {
+    #[cfg(not(feature = "async"))]
```
Burning1020, line 149: @mxpv `async` feature defined in `Cargo.toml` used here

crates/runc-shim/Cargo.toml
Burning1020: distinguish from "sync" feature. There would be two same funcs if removed.

crates/shim/src/io.rs
```diff
@@ -0,0 +1,289 @@
...
+use nix::sys::termios::Termios;
+
+use crate::error::{Error, Result};
+use crate::util::IntoOption;
+
+pub trait WriteCloser: Write {
+    fn close(&self);
+}
+
+pub trait IO: Sync + Send {
```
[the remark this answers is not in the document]
Burning1020: Done

crates/shim/src/io.rs
Burning1020: Done

crates/shim/src/lib.rs
Burning1020: Done, quite simple

crates/shim/src/monitor.rs
Burning1020: Done

crates/shim/src/monitor.rs
Burning1020: Done

Burning1020:
update done. ping @mxpv

mxpv (approved):
LGTM

mxpv:
@jiangliu @kzys PTAL?

mxpv:
Should we move `runc` dependent bits from `shim` crate to `runc-shim` ? (container, task).
Not a blocker for this PR, I can follow up on this.

fuweid (approved):
LGTM

kzys (approved):

Burning1020:
ping @mxpv

Burning1020:
thanks for reviewing!  @mxpv @fuweid @kzys

PR #28, opened 2022-02-13
[runc] Add async feature

mxpv:
Add `async` feature to isolate dependencies.

This will allow to exclude unnecessary dependencies (like don't pull `tokio` if not using async rust)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,11 +16,12 @@ jobs:
       - run: cargo check --examples --tests --all-targets
       - run: cargo fmt --all -- --check --files-with-diff
       - run: cargo clippy --all-targets --all-features -- -D warnings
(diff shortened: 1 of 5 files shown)


jiangliu:
How about grouping sync/async of lib.rs related code into submodules? Such as
```
pub use executor::*;

#[cfg(feature = "async")]
mod executor {
    use super::*;
    pub type Command = tokio::process::Command;
    impl RunC { ...  }
}


#[cfg(not(feature = "async"))]
mod executor {
    use super::*;
    pub type Command = std::process::Command;
    impl RunC { ...  }
}
```

But this way, we may have much less conditional compilation:)

mxpv:
Thanks, I've updated PR to reduce conditional attributes.
Instead of `executor` module I just have a couple different `impl Runc` blocks depending on current feature. Same trick with tests.

jiangliu (approved):

PR #29, opened 2022-02-13
Enhancements to the runc crate

jiangliu:
Enhancements to the runc crate
1) add unit test cases
2) add helpers to reduce duplicated code
3) avoid possible panic
4) refine inline documentation

### DIFF
--- a/crates/runc/src/io.rs
+++ b/crates/runc/src/io.rs
@@ -121,7 +121,7 @@ pub struct PipedIo {
 }
 
 impl PipedIo {
-    pub fn new(uid: u32, gid: u32, opts: IOOption) -> std::io::Result<Self> {
+    pub fn new(uid: u32, gid: u32, opts: &IOOption) -> std::io::Result<Self> {
         Ok(Self {
             stdin: Self::create_pipe(uid, gid, opts.open_stdin, true)?,
             stdout: Self::create_pipe(uid, gid, opts.open_stdout, false)?,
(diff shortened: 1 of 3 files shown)


mxpv:
Sorry, I've caused merge conflicts by merging https://github.com/containerd/rust-extensions/pull/28. Could you please rebase?

jiangliu:
> Sorry, I've caused merge conflicts by merging #28. Could you please rebase?

@mxpv I have rebased to the new base, but the new unit test case `io::tests::test_create_piped_io` fails on macos though succeeds on linux. And I have no macos system for local test, could you please help to take a look? 
I have temporarily disabled the unit test case for macos.

jiangliu:
> > Sorry, I've caused merge conflicts by merging #28. Could you please rebase?
> 
> @mxpv I have rebased to the new base, but the new unit test case `io::tests::test_create_piped_io` fails on macos though succeeds on linux. And I have no macos system for local test, could you please help to take a look? I have temporarily disabled the unit test case for macos.

And the exact failure point is PipedIo::create_pipe()
``` rust
        if let Some(f) = guard.as_ref() {
            let uid = Some(nix::unistd::Uid::from_raw(uid));
            let gid = Some(nix::unistd::Gid::from_raw(gid));
            nix::unistd::fchown(f.as_raw_fd(), uid, gid)?;
        }
```
Something gets wrong with the `nix::unistd::fchown(f.as_raw_fd(), uid, gid)`. I have also checked the arguments to to fchown on macos, it seems reasonable.

crates/runc/src/monitor.rs
```diff
@@ -72,9 +81,87 @@ impl DefaultMonitor {
...
+/// Execute a `Command` and collect exit status, output and error messages.
+///
+/// To collect output and error messages, pipes must be used for Command's stdout and stderr.
+/// This method will create pipes and overwrite stdout/stderr of `cmd`.
+///
+/// Note: invalid UTF-8 characters in output and error messages will be replaced with the `�` char.
+pub async fn execute<T: ProcessMonitor + Send + Sync>(
+    monitor: &T,
+    mut cmd: Command,
+) -> Result<(Exit, ExitStatus, String, String), Error> {
```
mxpv, line 101: May be introduce a struct? Too many return params.

crates/runc/src/io.rs
```diff
@@ -261,9 +239,82 @@ impl Io for NullIo {
...
+            open_stderr: false,
+        };
+        let io = PipedIo::new(1000, 1000, &opts).unwrap();
+
+        assert!(io.stdin().is_none());
+        assert!(io.stdout().is_none());
+        assert!(io.stderr().is_none());
+    }
+
+    #[cfg(not(target_os = "macos"))]
```
mxpv, line 268: I don't think it makes too much sense to target non Linux platforms, since runc is Linux specific. Lets just do 

```rust
#[cfg(test)]
#[cfg(target_os = "linux")]
mod tests {
```

crates/runc/src/io.rs
```diff
@@ -261,9 +239,82 @@ impl Io for NullIo {
         Ok(())
     }
 
-    /// closing only write side (should be stdout/err "from" runc process)
```
mxpv, line 264: Not directly related to these changes, but do we need `close_after_start` ?
(Rust generally does Drop instead of Close).
Should we clean this as well?

crates/runc/src/monitor.rs
jiangliu: Fixed

crates/runc/src/io.rs
jiangliu: fixed

crates/runc/src/io.rs
jiangliu: I feel it's still needed:)
Actually it reminds me that we need `impl<T: Io> Io for Arc<T>`.
The caller prepares an Arc<Io> object and passes a clone of it to Runc::create().
Runc::create() should call close_after_start() to release the writer side.
The caller could receive data from the Arc<Io> object then.

crates/runc/src/io.rs
```diff
@@ -55,6 +54,28 @@ impl Debug for dyn Io {
     }
 }
 
+impl<T: Io> Io for Arc<T> {
```
mxpv, line 57: I didn't get why do we need this. 

`Arc<T> automatically dereferences to T (via the Deref trait), so you can call T’s methods on a value of type Arc<T>.`

This seems redundant?

mxpv (approved):

mxpv:
Let's have this PR in. Will follow up fixes.

crates/runc/src/io.rs
mxpv: https://github.com/containerd/rust-extensions/pull/32

PR #35, opened 2022-02-21
Remove runc dependency from shim crate

mxpv:
Remove runc dependency from `shim` crate (move to `shim-runc` crate).

shim crate expected to be implementation independent.
(though i'm sure we could extract some parts back to `shim` in future)

PTAL @jiangliu @Burning1020 

### DIFF
--- a/crates/runc-shim/Cargo.toml
+++ b/crates/runc-shim/Cargo.toml
@@ -13,11 +13,14 @@ homepage = "https://containerd.io"
 async = []
 
 [dependencies]
+log = "0.4"
 nix = "0.23.1"
+libc = "0.2.95"
 time = { version = "0.3.7", features = ["serde", "std"] }
 serde = { version = "1.0.133", features = ["derive"] }
 serde_json = "1.0.74"
 oci-spec = "0.5.4"
+crossbeam = "0.8.1"
 
 containerd-shim = { path = "../shim", version = "0.2.0" }
 runc = { path = "../runc", version = "0.1.0" }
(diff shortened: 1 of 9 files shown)


Burning1020:
They do have some common parts indeed, as I see, the pain points is nothing more than `Io` trait defined in `runc` crate and used in `shim` crate. I suggest the IO trait shoud be treated as a common thing, we can only move IO to `shim` crate and give the caller one more function param as a callback function that can `set` io to `command`. In this way, `runc` crate could seperate from `shim` crate completely.

Burning1020:
1. Add a new function param in order to set command io, and call `io.close_after_start();` after `create` caller success. Same for other commands:
https://github.com/containerd/rust-extensions/blob/main/crates/runc/src/lib.rs#L172
```rust
    pub fn create<P>(
        &self,
        id: &str,
        bundle: P,
        opts: Option<&CreateOpts>,
        callback: fn(&mut command) -> Result<()>,
    ) -> Result<Response>
    where
        P: AsRef<Path>,
    {
        ...
        match opts {
            Some(CreateOpts { io: Some(io), .. }) => {
                callback(&mut cmd).map_err(|e| Error::IoSet(e.to_string()))?;
                let res = self.launch(cmd, true)?;
                Ok(res)
            }
```

2. Move out `Console` and `ConsoleSocket` from `runc` crate.
https://github.com/containerd/rust-extensions/blob/main/crates/runc/src/console.rs

How about this?

jiangliu:
Yes, the shim should be on the bottom of the dependency graph, and it should be neutral. It would be great to keep runc specific/related in shim-runc/runc. And late if we found common code in multiple shim implementation, we could then consolidate them into the shim crate.

jiangliu (approved):

mxpv:
I think its fine to keep Io with runc crate for now.
Would be curious to see alternative runtime implementations to see what common parts we could extract.
Let me proceed with this one, so I can publish a release.
We can follow up with more refactorings.

PR #36, opened 2022-02-22
shim-proto: group common proto types into types

jiangliu:
Group generated soruce files for common protobuf source files into
src/types, also simplify the build.rs.

Signed-off-by: Liu Jiang <gerry@linux.alibaba.com>
Signed-off-by: Quanwei Zhou <quanweiZhou@linux.alibaba.com>


crates/shim-protos/src/shim/empty.rs
```diff
@@ -1,157 +1,3 @@
-// This file is generated by rust-protobuf 2.27.1. Do not edit
-// @generated
+/// Reexport types::empty::* to keep backward compatibility.
```
mxpv, line 1: Do we need to keep backward compatibility? Until we reach 1.0,
we can just publish another major release with these changes.

jiangliu: The comments is a little misleading:(
It's here actually to avoid manually editing the auto-generated files.
Otherwise we need add add code into build.rs to change the reference paths for data types defined in types/mod.rs.
For example, we need to change `super::mount::Mount` to `crate::types::mount::Mount`.
And the reference path becomes longer, then it may break `cargo fmt`.

mxpv (approved):

PR #40, opened 2022-02-24
shim: Add async feature

abel-von:
added the async feature in shim,in which we added the async traits and async helper functions, and an empty async shim example provided.

### DIFF
--- a/crates/shim/src/asynchronous/mod.rs
+++ b/crates/shim/src/asynchronous/mod.rs
@@ -1,0 +1,413 @@
+use std::os::unix::fs::FileTypeExt;
+use std::os::unix::io::AsRawFd;
+use std::os::unix::net::UnixListener;
+use std::path::Path;
+use std::process::Command;
+use std::process::Stdio;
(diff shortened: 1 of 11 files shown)


crates/shim/src/asynchronous/mod.rs
```diff
@@ -0,0 +1,414 @@
+#![cfg(feature = "async")]
```
jiangliu, line 1: Is line 1 redundant?

abel-von: yes, I will remove it.

abel-von: removed

abel-von:
All the check errors are fixed, and I added a cargo check of all features in ci.yml. PTAL @jiangliu @mxpv @fuweid

abel-von:
I am working on the async runc-shim based on this PR, and I think it will be ready in these two days.

abel-von:
included in PR #50

PR #41, opened 2022-02-24
Rework exit signals and task creation

mxpv:
- `ExitSignal` is no longer `Arc`. Don't add extra overhead by default, its up to runtime implementations whether they need to count references. 
- Renames `get_task_service` to `create_task_service` to highlight that we expect task service to be created from there (and don't do clonable objects).
- Create `ShimTask` from `create_task_service` to avoid extra creation during `shim --start` call.
- Replaces `Arc<Mutex<Option<ShutdownType>>>` with `Once`. 


Burning1020:
lgtm

jiangliu (approved):

PR #42, opened 2022-02-26
runc-shim: implement stats and update task

Burning1020:
implement stats container task and update container task

### DIFF
--- a/crates/runc/src/lib.rs
+++ b/crates/runc/src/lib.rs
@@ -36,12 +36,10 @@
 //! A crate for consuming the runc binary in your Rust applications, similar to
 //! [go-runc](https://github.com/containerd/go-runc) for Go.
 use std::fmt::{self, Display};
-#[cfg(not(feature = "async"))]
-use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::{ExitStatus, Stdio};
 
-use oci_spec::runtime::{Linux, Process};
+use oci_spec::runtime::{LinuxResources, Process};
 
 pub mod console;
 pub mod container;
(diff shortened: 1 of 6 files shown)


Burning1020:
Well you know, launching runc binary is kind of heavy for tasks that were called frequently like _stats_ and _update_, thus I prefer to go straight to edit the cgroup paths of container process as what runc actually does.

crates/shim/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde = "1.0.136"
 uuid = { version = "0.8.2", features = ["v4"] }
 signal-hook = "0.3.13"
 oci-spec = "0.5.4"
+cgroups-rs = "0.2.8"
```
mxpv, line 27: `shim` is generic trait, we should not depend on linux specific concepts (like cgroups)

crates/runc/src/lib.rs
```diff
@@ -340,11 +339,16 @@ impl Runc {
...
-        std::fs::write(&filename, spec_json).map_err(Error::SpecFileCreationFailed)?;
+    pub fn update(&self, id: &str, resources: &LinuxResources) -> Result<()> {
+        let (mut temp_file, filename) = utils::make_temp_file_in_runtime_dir()?;
+        {
+            let f = temp_file.as_file_mut();
+            let spec_json =
+                serde_json::to_string(resources).map_err(Error::JsonDeserializationFailed)?;
+            f.write(spec_json.as_bytes())
+                .map_err(Error::SpecFileCreationFailed)?;
+            f.flush().map_err(Error::SpecFileCreationFailed)?;
```
mxpv, line 350: Should we extract to helper func for writing to a temp file?

mxpv (commented):
`runc-shim` needs a few `target_os = "linux"` checks to make CI happy

crates/shim/Cargo.toml
Burning1020: OK

crates/runc/src/lib.rs
Burning1020: Solved

mxpv (approved):

PR #46, opened 2022-03-01
runc-shim: impl close io and pids task

Burning1020:
Implements of close io and pids task.

### DIFF
--- a/crates/runc-shim/src/task.rs
+++ b/crates/runc-shim/src/task.rs
@@ -178,6 +178,11 @@ where
         Ok(Empty::new())
     }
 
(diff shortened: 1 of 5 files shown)


crates/runc-shim/src/task.rs
```diff
@@ -178,6 +189,11 @@ where
         Ok(Empty::new())
     }
 
+    fn close_io(&self, _ctx: &TtrpcContext, _req: CloseIORequest) -> TtrpcResult<Empty> {
+        // unnecessary close io here since fd was closed automatically after object was destroyed.
+        Ok(Empty::new())
```
Burning1020, line 194: `FileDesc` inside the `File` has implemented `Drop` trait which will close fd after object was destroyed. As a result, I think `close_io` task is unnecessary.

mxpv (approved):

PR #48, opened 2022-03-02
runc-shim: support shim cgroup and oom score

Burning1020:

### DIFF
--- a/crates/runc-shim/src/service.rs
+++ b/crates/runc-shim/src/service.rs
@@ -18,10 +18,11 @@
 
 use std::env::current_dir;
 use std::path::Path;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 use containerd_shim as shim;
 
+use runc::options::{DeleteOpts, GlobalOpts, DEFAULT_COMMAND};
 use shim::api::*;
 use shim::error::{Error, Result};
 use shim::monitor::{monitor_subscribe, Subject, Subscription, Topic};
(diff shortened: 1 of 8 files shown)


crates/runc-shim/src/service.rs
```diff
@@ -81,7 +88,33 @@ impl Shim for Service {
             None => {}
         }
 
-        let address = spawn(opts, &grouping, Vec::new())?;
+        let (child_id, address) = spawn(opts, &grouping, Vec::new())?;
+
+        #[cfg(target_os = "linux")]
```
jiangliu, line 93: How about split out this as a function?

crates/runc-shim/src/service.rs
```diff
@@ -144,6 +177,34 @@ impl Shim for Service {
     }
 }
 
+/// Sets the OOM score for the process to the parents OOM score + 1
```
jiangliu, line 180: All these functions may be grouped as a submodule, so we could reduce conditional compilation.

crates/runc-shim/src/service.rs
Burning1020: I was planning to do that, but worried about changing too much. Now that you have the same feeling , I'd like to try again :)

crates/runc-shim/src/service.rs
Burning1020: You're right. Should I mark `cgroup` as a new custom feature to replace `target_os = "linux"` at the same time?

Burning1020:
updated. ping @jiangliu

mxpv (approved):

PR #49, opened 2022-03-03
shim: refine set subreaper

Burning1020:
It's better to get rid of invoking "unsafe" libc function directly in `set_subreaper()` though we know it's quite safe in fact.


Burning1020:
My purpose is that we should try best not invoking unsafe functions directly. It can be replaced by other third-party crate and maintained by them.

Burning1020:
If you also agree with me, I'm going to open an issue to trace this problem for log time.

mxpv (approved):

PR #37, opened 2022-02-22
Add containerd integration tests

mxpv:
Enable integration test for runc-shim to make sure we're compliant with go version.

Fixes: https://github.com/containerd/rust-extensions/issues/6

Signed-off-by: Maksym Pavlenko <pavlenko.maksym@gmail.com>

PR #50, opened 2022-03-03
[runc-shim] support of async runc shim

abel-von:
Hi, I have finished the async codes of runc shim, and it can be compiled to a runnable conainerd-shim-runc-v2,  I have tested the crictl create/start/exec/stop, it behaves well.

I also refined the codes,  simplified the code architecture. make it much cleaner.

Now, for async codes, if someone want to construct a new shim binary of a new runtime, he or she just need to do the following things:
1. implement `CotainerFactory`, which is a Factory for creating container;
2. implement `ProcessLifecycle` for container init process,  `ProcessLifecycle` is a trait defines the process lifecycle actions, like start/kill/delete ;
3. implement `ProcessLifecycle` for container exec process, if exec process and init process has no difference, just implement the trait once;
4. implement `ProcessFactory` for exec processes;
5. assemble  these implementions to a `TaskService` and implements the `Service`;
6. provide async main function;

the details are in the codes in runc-shim/src/asynchronous.

I only refactored the codes in async feature, but I think the non async codes can also be refatored refer to the async codes.

this PR is based on the PR #40,  should I close the PR #40, only leave this PR?

I think it is a big improvement, could you please take a look? @mxpv @jiangliu @fuweid @Burning1020


Burning1020:
Awesome!

mxpv:
Should we keep just one version of shim implementation? (either sync or async).
It makes sense to keep and maintain sync/async versions for libraries, because we're not aware how it's going to be used by customers. But for runc shim we produce a binary and it would be challenging to support two versions of the same code.

abel-von:
Agree, but I'm not sure which one is the better choice, maybe we can keep these two versions of runc as a standard implementation of our shim lib for both sync and async users?

Burning1020:
How about taking "sync" as stable beta version and "async" as experimental alpha version? We can keep they two at the previous release, and the async version could be positive in somedays.

crates/runc/src/lib.rs
```diff
@@ -346,31 +340,54 @@ impl Runc {
...
+pub trait Executor {
+    fn execute(&self, cmd: Command) -> Result<(ExitStatus, u32, String, String)>;
+}
+
+#[cfg(feature = "async")]
+#[async_trait]
+pub trait Executor {
+    async fn execute(&self, cmd: Command) -> Result<(ExitStatus, u32, String, String)>;
+}
+
```
mxpv, line 370: I would prefer to have two `execute` implementations depending on `async` feature toggle. There is no need to have `Executor` trait here (moreover `async_trait` adds additional overhead to this).

crates/runc/src/lib.rs
```diff
@@ -749,25 +789,26 @@ mod tests {
...
 #[cfg(all(target_os = "linux", feature = "async"))]
 mod tests {
+    use std::sync::Arc;
+
     use super::io::{InheritedStdIo, PipedStdIo};
     use super::*;
-    use std::sync::Arc;
 
-    fn ok_client() -> Runc {
+    fn ok_client() -> Runc<DefaultExecutor> {
```
mxpv, line 797: Let's keep `Runc` with different APIs depending on `async` toggle (see comment above)

crates/runc/src/options.rs
```diff
@@ -113,7 +113,17 @@ pub struct GlobalOpts {
...
+            command: None,
+            debug: false,
+            log: None,
+            log_format: Default::default(),
+            root: None,
+            rootless: None,
+            set_pgid: false,
+            systemd_cgroup: false,
+            timeout: Default::default(),
+        }
```
mxpv, line 126: Why not `Default::default()` ?

crates/shim/src/asynchronous/mod.rs
```diff
@@ -0,0 +1,429 @@
...
+    let publisher = publisher::RemotePublisher::new(&ttrpc_address).await?;
+
+    // Create shim instance
+    let mut config = opts.unwrap_or_else(Config::default);
+
+    // Setup signals
+    let signals = setup_signals_tokio(&config);
+
+    if !config.no_sub_reaper {
+        asyncify(|| -> Result<()> { reap::set_subreaper() }).await?;
```
mxpv, line 125: Why not just `reap::set_subreaper()`? I'd use async only where its applicable.

crates/shim/src/common.rs
```diff
@@ -0,0 +1 @@
+pub type ShutdownFn = Box<dyn FnOnce() + Send + Sync>;
```
mxpv, line 1: Could you pls align these changes with changes from https://github.com/containerd/rust-extensions/pull/41

crates/shim/src/lib.rs
```diff
@@ -71,13 +72,19 @@ pub mod api {
...
 mod args;
+pub mod common;
 mod logger;
 pub mod monitor;
 pub mod mount;
 mod publisher;
 mod reap;
 pub mod util;
 
+#[cfg(feature = "async")]
```
mxpv, line 83: Can we do it mutually exclusive? e.g. don't allow sync API when `async` enabled (I think its ok to define async api side by side with sync ones and use feature toggle to exclude).

crates/shim/src/asynchronous/task.rs
```diff
@@ -0,0 +1,212 @@
...
+};
+use crate::asynchronous::container::{Container, ContainerFactory};
+use crate::common::ShutdownFn;
+use crate::util::AsOption;
+use crate::TtrpcResult;
+
+/// TaskService is a Task template struct, it is considered a helper struct,
+/// which has already implemented `Task` trait, so that users can make it the type `T`
+/// parameter of `Service`, and implements their own `ContainerFactory` and `Container`.
+pub struct TaskService<F, C> {
```
mxpv, line 28: Can we move this to `runc-shim` to align with sync APIs.
See here https://github.com/containerd/rust-extensions/pull/35 for context.

crates/shim/src/asynchronous/io.rs
```diff
@@ -0,0 +1 @@
+
```
mxpv, line 1: Not needed?

crates/runc/src/lib.rs
```diff
@@ -945,3 +986,38 @@ mod tests {
         assert!(!response.output.is_empty());
     }
 }
+
+#[derive(Clone)]
+pub struct DefaultExecutor {}
```
mxpv, line 991: All custom executors will depend on Tokio anyway, so this can be omitted (see comments above).

crates/runc-shim/src/console.rs
```diff
@@ -0,0 +1,43 @@
+use std::os::unix::io::RawFd;
```
mxpv, line 1: Could you pls add the copyright header to all new files.

crates/runc-shim/Cargo.toml
```diff
@@ -10,7 +10,7 @@ description = "Rust implementation of containerd's runc v2 shim runtime"
 homepage = "https://containerd.io"
 
 [features]
-async = []
+async = ["containerd-shim/async","runc/async", "tokio", "futures", "async-trait"]
```
mxpv, line 13: Nit: needs space between `","`

crates/runc/src/lib.rs
abel-von: Actually the reason I add Executor trait here is not because the Command is different from sync to async. It is actually because the `wait` function of `Command` may block because runc-shim may start a child reaper, which will catch the SIGCHLD signal, that makes `wait` function can not receive the SIGCHLD, and wait there forever. As the runc crate can not depends on the shim crate, I can only expose a `Executor` trait in runc crate, and let runc-shim to implements their own `execute` logic based on the monitor from shim. the details is in implementation for the `ShimExecutor`. 
And I think the Monitor should be a helper struct for all shim implementation, so I think it's better if it is in the shim crate.

Burning1020: Same problem occurs in Sync version too. Most of [integration test](https://github.com/containerd/rust-extensions/pull/37#issuecomment-1059849843) failed with `Others("Other: unknown error after kill: error occured in runc: no child processes (os error 10)"): unknown` may caused by runc-shim process have set child subreaper. I also have found that everything's ok on CentOS but wrong with Ubuntu. So I think we still need to keep this `Executor` in Sync version.

Do we have a better solution?

crates/runc/src/lib.rs
abel-von: Maybe we can not make it,  as refer to the answer of the comment above.

crates/runc/src/options.rs
abel-von: changed

crates/shim/src/asynchronous/mod.rs
abel-von: I  thought the set_subreaper may be a blocking operation.  Maybe it is not? anyway, I followed your instruction to changed it to call `reap::set_subreaper()` directly

crates/shim/src/common.rs
abel-von: changed

crates/shim/src/lib.rs
abel-von: Now the async and sync specific codes are splited into `synchronous` and `asynchronous` folder `shim` and `runc-shim`. pls check

crates/shim/src/asynchronous/task.rs
abel-von: I implemented `Task` trait in `shim` crate with this `TashService` template, and the codes in the implemetation are not runtime specific, the `runc-shim` (or any other shim like `kata-shim`, `gvisor-shim` in the future?) do not need to implement the `Task` trait anymore, they just implements the trait exposed in `shim` crate like `ContainerFactory`, `ProcessLifecycle`. and construct their implementation of `Task` trait with those traits implementions.

abel-von: Actually we have an implementation of kata-shim, if every kata-shim implements the `Task` trait, there will be many duplicate codes.

crates/shim/src/asynchronous/io.rs
abel-von: removed

crates/runc/src/lib.rs
abel-von: please refer to the answer of the first comment.

crates/runc-shim/src/console.rs
abel-von: added

crates/runc-shim/Cargo.toml
abel-von: changed

crates/runc/src/lib.rs
mxpv: I see what you're trying to do.
You want to inject custom logic before/after spawning the process.
It might cleaner to allow extensions via runc's Builder.

Something to inspire from Tokio:
https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/builder.rs#L68
https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/builder.rs#L351

```rust
runc::builder().custom_spawner(...) ?
```

WDT?

mxpv (approved):

PR #55, opened 2022-03-08
runc-shim: support remote publisher and extra some common methods

Burning1020:
Allow shim publish task event to containerd, and also extra some common convert methods


Burning1020:
Merrge 57, 58 first.

@mxpv With PR 57, 58 and 37, we can make containerd integration tests happy~

PR #58, opened 2022-03-09
bugfix: return a tmp dir when 'XDG_RUNTIME_DIR' was not set

Burning1020:
use a tmp dir when 'XDG_RUNTIME_DIR' was not set


Burning1020:
merge #57 first

mxpv (approved):

PR #60, opened 2022-03-14
Fix tokio not exit and signal handle exit issue

### LINKED ISSUE #59: [bug][async] shim process not exit after remove the container
After we do `crictl exec -ti $container_id bash` and exit, and then `crictl rmp -f $podid`, the shim process do not exit. this is because the spawned io copy task will make a blocking stdin read task in the tokio runtime, when main exit, it will call drop of tokio runtime, which will wait for the exit of that task, then blocking main thread exit.

and there is another bug, when we `kill <pid of shim>` when container is running, the shim will not exit, which is expected, but the subsequent exec can not exit, that is because the signal handler thread exit when SIGTERM received, for the subsequent signals, including SIGCHLD has no one to handle them.

abel-von:
fix #59

### DIFF
--- a/crates/runc-shim/src/main.rs
+++ b/crates/runc-shim/src/main.rs
@@ -26,11 +26,18 @@ fn main() {
 }
 
 #[cfg(feature = "async")]
-#[tokio::main]
-async fn main() {
-    containerd_shim::asynchronous::run::<crate::asynchronous::Service>(
-        "io.containerd.runc.v2-rs",
-        None,
-    )
-    .await;
+fn main() {
+    let body = async {
+        containerd_shim::asynchronous::run::<crate::asynchronous::Service>(
+            "io.containerd.runc.v2-rs",
+            None,
+        )
+        .await;
+    };
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .expect("Failed building the runtime");
+    runtime.block_on(body);
+    runtime.shutdown_background();
 }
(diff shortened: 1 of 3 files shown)


mxpv:
According to https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_background

> Shuts down the runtime, without waiting for any spawned tasks to shutdown.

I'd prefer to avoid use of it. Possibly we could perform the io copy with timeout or properly shutdown background tasks instead?

mxpv:
According to https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_background

> Shuts down the runtime, without waiting for any spawned tasks to shutdown.

I'd prefer to avoid use of it. Possibly we could perform the io copy with timeout or properly shutdown background tasks instead?

abel-von:
I think it is hard to determine the timeout value for the io copy, because the stdin or stdout may not receive anything for a long time. the background task is a blocking task and it is blocked in the `pipe_wait` function in libc, only closing the write half of the pipe can we break the blocking wait.
Besides, if we don't call `shutdown_background` before main exit, then any blocking task may prevent the shim process from exiting, and make a residual process. I think it is hard to list all the situations, and make sure 100 percent that all tasks can properly shutdown.

mxpv:
I understand the intent, but using `shutdown_background` essentially means to ignore `Drop`, so other things may not shutdown properly. This seems a bit aggressive.

> I think it is hard to determine the timeout value for the io copy, because the stdin or stdout may not receive anything for a long time.

Perhaps we could isolate this with nested runtime?

```rust
runtime.block_on(async move {
       let inner_runtime = Runtime::new().unwrap();
       // spawn blocking IO stuff.
       inner_runtime.shutdown_background();
   });
```

abel-von:
I have to close this PR, because if I update and push the codes again, the CI check will fail because of the `path-isolate` update issue I mentioned in the PR #61, the commit is included in the #61,  Maybe we can continue our discussion about the shutdown issue in PR #61?

PR #61, opened 2022-03-18
[async] pass all integration tests

abel-von:
All the task ttrpc call implemented in synchronous are also implemented in async feature, and all the containerd integration tests passed, please check.


abel-von:
I checked the CI error
```
  Downloaded path-dedot v3.0.16
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/path-dedot-3.0.16/Cargo.toml`

Caused by:
  feature `edition2021` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["edition2021"]` to enable this feature
Error: Process completed with exit code 101.
```

I think it is because the path-dedot crate published a new version and makes it a requirement of edition 2021. 
I checked where does this crate dependency come from with `cargo tree`, and found that it is a indirect denpendency, from `path-absolute` crate, which is dependent by `runc` crate.
then I checked the codes,  and found that we use this crate to get the absolute path from any path like relative path or symlink, I think the `std::fs::canonicalize` can do this, so I will remove the dependency of `path-absolute`;

abel-von:
After I changed the call to `std::fs::canonicalize`,  the unit test failed, because  `std::fs::canonicalize` need the existence of the path, which we can not make sure.
I think we have two options:
1. use cargo `cargo update -p path-dedot --precise 3.0.12` to pin the dependency to 3.0.12. and upload Cargo.lock to repository.
2. specify the edition to 2021 in our Cargo.toml.

I don't know which is better. so what do you think about? @mxpv @Burning1020

abel-von:
`edition2021` has been stabilized in the 1.56 release, so update the rust-toolchain.toml to update the cargo version to 1.56 resolve this error.

abel-von:
added the logic to close the stdin pipe when delete the container, that fix #60. and removed the call of `shutdown_background`.  please check @mxpv

mxpv:
Should we also run async version of the runtime on CI too? (just add another step `Integration (Async)` to https://github.com/containerd/rust-extensions/blob/main/.github/workflows/ci.yml

abel-von:
added steps of integration test for asycn shim.   PTAL @mxpv

mxpv (approved):

PR #74, opened 2022-05-03
Collect build timings

mxpv:
Rust 1.60 now allows collecting build timings.
This PR updates `ci.yml` to collect a timing report.
The report will be published as an HTML artifact on Github Actions.
This information would be useful to understand how various dependencies affect build times as well as better understand how build times change over time.

More info: https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#cargo---timings

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.59"
+channel = "1.60"
 components = ["rustfmt", "clippy"]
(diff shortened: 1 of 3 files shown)


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.59"
+channel = "1.60"
```
Yuna-Tomi, line 2: I found that MSRV notification on `README.md` is still remaining 1.54.
It seems to be better to update it too 🙏

Burning1020:
Nice. LGTM

PR #79, opened 2022-05-17
runc-shim: Update console copy

Burning1020:
Recently I've found the bash doesn't display as expected sometimes when do exec task with a tty. Specificaly, if you type an alphabet from keyboard, the bash will print that alphabet as we expected. But in this bad case, nothing is printed untill you type an alphabet again. 

This only happens after [this rust-lang commit](https://github.com/rust-lang/rust/commit/16236470c1774f88374bab29d2b9d1875cb97246) in v1.50 that uses the efficient [splice](https://man7.org/linux/man-pages/man2/splice.2.html) to move data frrom a tty to a pipe with host kernel under 4.15 . Here is a simple diagram to show the tty intereaction in our case:
```bash
  containerd      containerd-shim-runc-v2-rs
	↑                       |
	| Read(1)               |
	|                  Splice(2)
   /xxxxxxxx-stdout       <--------     /dev/ptmx    <------    /dev/pts/x
```
However, both the Read(1) and Splice(2) will lock the pipe and they are from a loop in different thread, so that there could be a data race among them. When the user have typed an alphabet , if Read(1) get the lock, the data will be read by containerd and returned to the exec bash, everything's ok, but if Splice(2) get the lock, it will hold this lock until the next input, blocking Read(1) thread so that no output will return to the bash.

This commit replaced the default `std::io::copy()` by a normal usermode `copy()` to avoid using splice. Happy to hear that @fuweid will report this bug to the kernel group and track this issue.

Note: This won't hanppen in async shim for that it uses a normal copy from tokio io. 

### DIFF
--- a/crates/runc-shim/src/synchronous/container.rs
+++ b/crates/runc-shim/src/synchronous/container.rs
@@ -38,7 +38,7 @@ use shim::{io_error, other, other_error};
 
 use crate::common::receive_socket;
 use crate::common::ProcessIO;
-use crate::synchronous::io::spawn_copy;
+use crate::synchronous::io::spawn_copy_for_tty;
 
 pub trait ContainerFactory<C> {
     fn create(&self, ns: &str, req: &CreateTaskRequest) -> Result<C>;
@@ -260,7 +260,7 @@ impl Process for CommonProcess {
                 .write(true)
                 .open(self.stdio.stdin.as_str())
                 .map_err(io_error!(e, "open stdin"))?;
-            spawn_copy(stdin, f, None, None);
+            spawn_copy_for_tty(stdin, f, None, None);
         }
 
         if !self.stdio.stdout.is_empty() {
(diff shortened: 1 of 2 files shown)


Burning1020:
Here is a simple program to reproduce it: https://gist.github.com/Burning1020/4d99bf1cc4934b44ab383a882d21637e.
We expect 'a' will be printed 1 time/s.

mxpv (approved):

zzyyzte:
Can you explain in more detail why split and read system calls generate lock competition and under what conditions?  Thanks very much.

Burning1020:
@zzyyzte This bug will happen on condition that one process read a named pipe and another process copy data from a tty to the pipe. On the one hand, read process will struck in a loop trying to read the pipe. On the other, write process also struck in a loop of splicing data. Go through the kernel code of [splice_file_to_pipe](https://elixir.bootlin.com/linux/latest/source/fs/splice.c#L1003), you could see splice will hold the pipe lock, move data to the pipe and then returns. After the first tty data moved to pipe, both two loop will competed in the lock. If read get it first, the byte will be read but if splice get it first, it will pass the second/third/... data until the read loop get the lock and read them up.

PR #95, opened 2022-09-21
Update shim dependencies

mxpv:
This PR updates shim dependencies (updates ttrpc to latest versions).
Since TTRPC is now using latest protobuf 3.1, there is some migration work.

Close https://github.com/containerd/rust-extensions/pull/93 https://github.com/containerd/rust-extensions/pull/92 https://github.com/containerd/rust-extensions/pull/91 https://github.com/containerd/rust-extensions/pull/82


crates/runc-shim/src/synchronous/container.rs
```diff
@@ -21,6 +21,7 @@ use std::os::unix::io::{AsRawFd, FromRawFd};
 use std::path::Path;
 use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
 
+use containerd_shim::protos::protobuf::EnumOrUnknown;
```
Burning1020, line 24: should we make a rule of referencing packages order, like std/3rd/self

crates/shim-protos/src/shim/mod.rs
```diff
@@ -13,3 +13,6 @@ pub mod shim_ttrpc_async;
 pub(crate) mod empty;
 pub(crate) mod mount;
 pub(crate) mod task;
+
+mod fieldpath;
+mod gogo;
```
Burning1020, line 18: public or not?

crates/shim/src/cgroup.rs
```diff
@@ -109,19 +109,19 @@ pub fn collect_metrics(pid: u32) -> Result<Metrics> {
         match sub_system {
             Subsystem::CpuAcct(cpuacct_ctr) => {
                 let mut cpu_usage = CPUUsage::new();
-                cpu_usage.set_total(cpuacct_ctr.cpuacct().usage);
+                cpu_usage.total = cpuacct_ctr.cpuacct().usage;
                 let mut cpu_stat = CPUStat::new();
-                cpu_stat.set_usage(cpu_usage);
-                metrics.set_cpu(cpu_stat);
+                cpu_stat.usage = Some(cpu_usage).into();
```
Burning1020, line 114: Does the new ttrpc not provide setter functions? I think `Some(xxx).into` is a little complex by writing.

mxpv: Protobuf 3 has removed setters/getters for generated structs.
`usage` has [MessageField](https://docs.rs/protobuf/latest/protobuf/struct.MessageField.html) type, which is a wrapper for `Option<Box<T>>` (replacement for `SingularPtrField` used in Protobuf 2). I found `Some().into()` to be the shortest syntax possible here.

crates/shim-protos/src/shim/mod.rs
mxpv: I plan to remove the generated code from the repo, so I don't have strong preference here.

crates/runc-shim/src/synchronous/container.rs
mxpv: I've fixed imports in https://github.com/containerd/rust-extensions/pull/95/commits/702d0212f4bf398c17a1d70c5bda28577fb19475
Can you elaborate on "make a rule"? I could not find the way to enforce imports order on CI :(

crates/shim/src/cgroup.rs
mxpv: It's also possible to use `MessageField::some()`, but it looks weird in the context of Rust with std `Option` type. So I'd prefer to use `Some` here.

Burning1020: that's OK

crates/runc-shim/src/synchronous/container.rs
Burning1020: Same with golang projects, I need to seperate these packages into three parts(std/3rd/self) by myself, the order of each part will be updated automatically by IDE or rust plugins. Thus, is it necessary to make demarcation? I also see many other repos they don't care about that.

crates/runc-shim/src/synchronous/container.rs
```diff
@@ -35,6 +34,7 @@ use shim::protos::cgroups::metrics::Metrics;
 use shim::util::{convert_to_timestamp, read_pid_from_file};
 use shim::Console;
 use shim::{io_error, other, other_error};
+use time::OffsetDateTime;
```
Burning1020, line 37: mark

crates/runc-shim/src/synchronous/container.rs
```diff
@@ -21,11 +21,10 @@ use std::os::unix::io::{AsRawFd, FromRawFd};
 use std::path::Path;
 use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
 
+use containerd_shim as shim;
+use containerd_shim::protos::protobuf::EnumOrUnknown;
```
Burning1020, line 25: mark

crates/runc-shim/src/synchronous/service.rs
```diff
@@ -21,6 +21,7 @@ use std::path::Path;
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::sync::Arc;
 
+use containerd_shim::protos::protobuf::MessageDyn;
```
Burning1020, line 24: mark

crates/shim/src/asynchronous/processes.rs
```diff
@@ -18,13 +18,14 @@ use std::os::unix::io::AsRawFd;
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use containerd_shim_protos::protobuf::EnumOrUnknown;
```
Burning1020, line 21: mark

crates/runc-shim/src/synchronous/container.rs
Burning1020: If it's troublesome, just let it go as it doesn't impact compile.

mxpv:
So I just found `generate_accessors` toggle that generated accessors. So I've reverted some code and got rid of these extra Enum imports. Unfortunately this does not make the diff a lot less, because accessors were renamed from `get_xyz()` to `xyz()`

fuweid (approved):

fuweid:
ping @Burning1020

Burning1020:
Lgtm

PR #96, opened 2022-09-23
Validate Rust docs on CI

mxpv:
This will validate documentation on CI (and makes sure rust doc doesn't fail).


Burning1020:
Lgtm

PR #97, opened 2022-09-25
Add Rust fmt rules to format imports

mxpv:
This is a follow up for discussion here: https://github.com/containerd/rust-extensions/pull/95#discussion_r978213442 whether its possible to format imports.
Actually turns out there is a way to enforce it, however the majority of config
items remain unstable, so we have to call cargo fmt with `+nightly` switch.

See new [rustfmt.toml](https://github.com/containerd/rust-extensions/blob/3e521ff3b53f50b6658efa4d91be63bd9f49b65c/rustfmt.toml)

@Burning1020 WDYT?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,9 +22,13 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - run: cargo check --examples --tests --all-targets
       - run: cargo check --examples --tests --all-targets --all-features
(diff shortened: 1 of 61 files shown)


Burning1020:
Thanks for your work @mxpv 
I agree to add these config items despit of its instability.

Burning1020 (commented):
Thanks for your work @mxpv
I agree to add these config items despit of its instability.

Burning1020 (approved):
lgtm

PR #98, opened 2022-09-27
Remove generated code from shim-protos (and fix CI)

mxpv:
While working on https://github.com/containerd/rust-extensions/pull/97 I found an interesting issue. Since we currently commit all generated code, we should pin protobuf version to be exactly `3.1.0`. Otherwise, whenever Protobuf crate gets updated, `shim-protos` will be broken with something like:
```
--> crates/shim-protos/src/cgroups/gogo.rs:26:49
[413](https://github.com/containerd/rust-extensions/actions/runs/3129846807/jobs/5079425017#step:4:414)
   |
[414](https://github.com/containerd/rust-extensions/actions/runs/3129846807/jobs/5079425017#step:4:415)
26 | const _PROTOBUF_VERSION_CHECK: () = ::protobuf::VERSION_3_1_0;
[415](https://github.com/containerd/rust-extensions/actions/runs/3129846807/jobs/5079425017#step:4:416)
   |                                                 ^^^^^^^^^^^^^ help: a constant with a similar name exists: `VERSION_3_2_0`
```

This is because, according to [semver](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), when we define `protobuf = "3.1.0"` in Cargo.toml, cargo may use any newer version up to < 2.0.0.

Another approach, more preferable in Rust, is to exclude any generated code from repo, and  use `include!` macro instead. See bindgen example: https://rust-lang.github.io/rust-bindgen/tutorial-4.html

The only thing to keep in mind about `include!` macro is that it does not support any form of `#![allow(XYZ)]` attributes, so we have to remote those to build time (and define at crate level) - https://github.com/rust-lang/rfcs/issues/752


Burning1020 (approved):
LGTM

PR #100, opened 2022-10-20
fix container process with stdout killed by sigpipe when containerd was stopped

zzyyzte:
fix container process with stdout killed by sigpipe when containerd was stopped

### DIFF
--- a/crates/runc-shim/src/common.rs
+++ b/crates/runc-shim/src/common.rs
@@ -20,7 +20,6 @@ use containerd_shim::{
     api::{ExecProcessRequest, Options},
     io::Stdio,
     io_error, other, other_error,
-    util::IntoOption,
     Error,
 };
 use log::{debug, warn};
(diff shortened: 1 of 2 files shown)


crates/runc-shim/src/common.rs
```diff
@@ -93,13 +92,15 @@ pub fn create_io(
...
-            stderr: stdio.stderr.to_string().none_if(|x| x.is_empty()),
+        let opts = IOOption {
+            open_stdin: !stdio.stdin.is_empty(),
+            open_stdout: !stdio.stdout.is_empty(),
+            open_stderr: !stdio.stderr.is_empty(),
         };
+        let io = PipedIo::new(_io_uid, _io_gid, &opts).unwrap();
+
         pio.io = Some(Arc::new(io));
-        pio.copy = false;
```
Burning1020, line 102: Instead of piping io, we have a different direct IO solution: https://github.com/containerd/rust-extensions/pull/24#issuecomment-1040619960.

Burning1020 (changes requested):

Burning1020:
Could you add some details like bug description, containerd log output?

mxpv:
ping @zzyyzte are you still interested in contributing this?

zzyyzte:
yes, but I can't remember exactly how this problem came about...

Burning1020:
@zzyyzte @mxpv I knew this problem exactly. This issus is caused by closing the "read" side of io fifo when containerd stopped. Read side should be opened in shim when is copy io is disabled. I'll fix this.

PR #111, opened 2022-12-28
runc-shim: fix leaking files for exec process

Burning1020:
A pid file under bundle path will be created after the exec process created, so it should also be deleted before process deleted.


Burning1020:
@mxpv conflicts solved. This two test, `TestTaskUpdate` and `TestContainerMetrics`, have been failed for a long time. They all with regard to some operations in Cgroup. But I did't figure it out.

mxpv (approved):

PR #126, opened 2023-03-15
Implement `list` function

wedsonaf:
This is needed by remote snapshotters: once they report "already exists", containerd tries to find the snapshot via `list`. If it's not implemented, the "already exists" trick to prevent layer download doesn't work.

This is still missing a filtering function, but allows remote snapshotters to work.

### DIFF
--- a/crates/snapshots/Cargo.toml
+++ b/crates/snapshots/Cargo.toml
@@ -19,6 +19,9 @@ prost-types = "0.11"
 tokio = { workspace = true, features = ["sync"] }
 tokio-stream = "0.1.8"
 serde = { version = "1.0", features = ["derive"] }
+async-stream = "0.3.3"
+futures = "0.3.17"
+pin-utils = "0.1.0"
 
 # tonic v0.8.1 depends on axum-core v0.2.2, which has security vulnerability:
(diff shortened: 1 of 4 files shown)


crates/snapshots/Cargo.toml
```diff
@@ -19,6 +19,9 @@ prost-types = "0.11"
 tokio = { workspace = true, features = ["sync"] }
 tokio-stream = "0.1.8"
 serde = { version = "1.0", features = ["derive"] }
+async-stream = "0.3.3"
+futures = "0.3.17"
```
mxpv, line 23:
```suggestion
futures.workspace = true
```

mxpv (commented):
One minor nit,
otherwise LGTM

wedsonaf:
> One minor nit, otherwise LGTM

Thanks for reviewing! I've folded your suggestion in.

After this is in, I only need one more one-liner (#130) to upstream all the changes I need to build the snapshotter in https://github.com/microsoft/kata-containers/tree/wedsonaf/tardev/src/tardev-snapshotter.

mxpv (approved):

PR #127, opened 2023-03-15
Make shim / runc-shim APIs more consistent

mxpv:
Current `shim` and `runc-shim` sync/async APIs are not consistent.

In current implementation, async `shim` crate keeps `container` and `task` modules, while sync versions of these modules are kept in `runc-shim` crate.

This PR moves these to `runc-shim` crate for consistency.

In longer run, I'd like to slimify `shim` crate more and remove all implementation details, so it handles shim lifecycle/events only.

image

### DIFF
--- a/crates/runc-shim/src/asynchronous/mod.rs
+++ b/crates/runc-shim/src/asynchronous/mod.rs
@@ -20,13 +20,10 @@ use ::runc::options::DeleteOpts;
 use async_trait::async_trait;
 use containerd_shim::{
     asynchronous::{
-        container::Container,
         monitor::{monitor_subscribe, monitor_unsubscribe, Subscription},
         processes::Process,
         publisher::RemotePublisher,
-        spawn,
-        task::TaskService,
-        ExitSignal, Shim,
+        spawn, ExitSignal, Shim,
     },
     event::Event,
     io_error,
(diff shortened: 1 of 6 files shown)


crates/runc-shim/src/asynchronous/mod.rs
```diff
@@ -45,7 +42,12 @@ use crate::{
     common::{create_runc, has_shared_pid_namespace, ShimExecutor, GROUP_LABELS},
 };
 
+mod container;
 mod runc;
+mod task;
+
+use container::Container;
+use task::TaskService;
```
Burning1020, line 50: Is it resonable to make `use container::Container;` and `use task::TaskService;` be in `use crate::{}`？Just to keep it clean. :)

Burning1020 (approved):
LGTM

crates/runc-shim/src/asynchronous/mod.rs
mxpv: Like `crate::asynchronous::container::Container` ?

crates/runc-shim/src/asynchronous/mod.rs
```diff
@@ -20,13 +20,10 @@ use ::runc::options::DeleteOpts;
 use async_trait::async_trait;
 use containerd_shim::{
     asynchronous::{
-        container::Container,
         monitor::{monitor_subscribe, monitor_unsubscribe, Subscription},
```
Burning1020, line 23: move them to here?

mxpv: On my list, but this requires slightly more refactoring.
Will follow up in another PR.

Issue #4, opened 2021-12-12
Will the rust-extensions support Windows?


jiangliu:
Seems currently implementation supports UNIX-like OSs only, is there any plan for Windows or it's ok to focus on Unix-like OSs?

mxpv:
I don't have windows environment, so I haven't tried it.
But it would be good to support all platforms though.
I think a good starting point would be to get [ttrpc-rust](https://github.com/containerd/ttrpc-rust) running on windows CI, to make sure nothing is broken.
Rust extensions should be straightforward to port

jiangliu:
It would be great to support Windows, but it does need some efforts. So how about marking it as an TODO?

mxpv:
Sure, let's keep this open

jsturtevant:
was able to get most of the shim related code to build using https://github.com/containerd/ttrpc-rust/issues/132 with a few tweaks

PR #139, opened 2023-04-26
Windows support for the synchronous shim

jsturtevant:
Fixes: #4

This adds Windows support for the synchronous shim on windows.  It builds on the support added to [ttrpc-rust for Windows](https://github.com/containerd/ttrpc-rust/pull/172).   It does require a couple updates (https://github.com/containerd/ttrpc-rust/pull/182 and https://github.com/containerd/ttrpc-rust/pull/185) and a release from that project.

Couple things to note:

- Windows doesn't have the same signal handling as Linux so this adds a real basic implementation  (thanks @danbugs for helping).  It looks like the Linux implementation uses SIGCHILD to handle child processes but Windows doesn't have a similar concept. If required down the line we could look into something as described in https://learn.microsoft.com/en-us/previous-versions/ms811896(v=msdn.10)#signals-and-signal-handling
- Windows shim logging is implemented using named Pipes in Containerd.  I've added detailed comments in the code on some assumptions.  The file logger does work for windows so I didn't completely remove it from compiling on windows, in case we wanted to make that configurable (it mostly makes debugging this crate easier).
- The windows shim activation is slightly different. Again I've left comments inline but the basic pattern comes from https://github.com/microsoft/hcsshim/blob/v0.10.0-rc.7/cmd/containerd-shim-runhcs-v1/serve.go#L57-L70. This seemed simpler than re-writing the rust `command` implementation in this repository but does add some complexity to how the shim boots.
- There are several crate in here that either won't ever work (runc) or I didn't focus on initially. This means the `cargo build` from root won't just work.  This means a user would need to call `cargo build` from the crate folders that do support windows (shim/shim-protos) 

Todo:

- [x] add a Windows pipeline to validate code
- [x] respond to initial feedback
- [x] get a release of ttrpc


crates/shim/src/synchronous/mod.rs
```diff
@@ -224,6 +281,22 @@ where
...
+    let mut server = Server::new();
+    let address = socket_address(&flags.address, &flags.namespace, &flags.id);
+    server = server.bind(address.as_str())?;
+    Ok(server)
+}
+
+#[cfg(unix)]
+fn create_server(_flags: args::Flags) -> Result<Server> {
+    let mut server = Server::new();
+    server = server.add_listener(SOCKET_FD)?;
```
mxpv, line 295: I'd prefer to have one function that works on both platforms, it's a bit easier to maintain (same applies for others)

```rust
fn create_server(_flags: args::Flags) -> Result<Server> {
     let mut server = Server::new();

     #[cfg(unix)]
     {
         server = server.add_listener(SOCKET_FD)?;
     }

     #[cfg(windows)]
     {
         ...
     }

     Ok(server)
 }
```

mxpv (commented):
This is awesome!
Would be nice to add windows target to our CI to run tests on Windows
(and fix clippy/rustfmt complains)

jsturtevant:
> Would be nice to add windows target to our CI to run tests on Windows

💯 one of the tasks I have to take care to get this out of draft. Thanks for the quick response!

crates/shim/src/synchronous/mod.rs
jsturtevant: I've converted most of these to a single function in https://github.com/containerd/rust-extensions/pull/139/commits/59f358fa1a799d786efc857f5f012d41dd65bd25

There were two places where I am not sure to handle it since the functions return different values:

```
#[cfg(windows)]
fn start_listener(address: &str) -> std::io::Result<()> {]

#[cfg(unix)]
fn start_listener(address: &str) -> std::io::Result<UnixListener> {
```

and the signal handling process stuff. 


Any pointers would be appreciated.

jsturtevant:
> > Would be nice to add windows target to our CI to run tests on Windows
> 
> 💯 one of the tasks I have to take care to get this out of draft. Thanks for the quick response!

I've added a job for Windows. Since it only supports a subset of the workspaces right now, it is is own job.  Also note until (https://github.com/containerd/ttrpc-rust/pull/182 and https://github.com/containerd/ttrpc-rust/pull/185) both merge the unit tests on windows are flaky.

jsturtevant:
the ubuntu 20.04 failure doesn't look related:

```
sudo -E PATH=$PATH make integration
  shell: /usr/bin/bash -e {0}
  env:
    GOPROXY: direct
    TEST_RUNTIME: io.containerd.runc.v2-rs
    TESTFLAGS_PARALLEL: 1
    EXTRA_TESTFLAGS: -no-criu -test.skip='(TestContainerPTY|TestContainerExecLargeOutputWithTTY|TestTaskUpdate|TestTaskResize)'
    TESTFLAGS_RACE: -race
+ integration
go: google.golang.org/genproto@v0.0.0-202[3](https://github.com/containerd/rust-extensions/actions/runs/4825275170/jobs/8595977056?pr=139#step:7:3)0306155012-7f2fa6fef1f[4](https://github.com/containerd/rust-extensions/actions/runs/4825275170/jobs/8595977056?pr=139#step:7:4) requires
	cloud.google.com/go/bigquery@v1.48.0 requires
	github.com/apache/arrow/go/v10@v10.0.1 requires
	gonum.org/v1/gonum@v0.11.0 requires
	golang.org/x/exp@v0.0.0-20191002040644-a13[5](https://github.com/containerd/rust-extensions/actions/runs/4825275170/jobs/8595977056?pr=139#step:7:5)5ae1e2c3 requires
	golang.org/x/mobile@v0.0.0-20190719004257-d2bd2a29d028: invalid version: git ls-remote -q origin in /home/runner/go/pkg/mod/cache/vcs/1ba9cdedc[6](https://github.com/containerd/rust-extensions/actions/runs/4825275170/jobs/8595977056?pr=139#step:7:6)63bd56a968[7](https://github.com/containerd/rust-extensions/actions/runs/4825275170/jobs/8595977056?pr=139#step:7:7)f5910f19e1ac5f7f472742441d370be599c409f[8](https://github.com/containerd/rust-extensions/actions/runs/4825275170/jobs/8595977056?pr=139#step:7:8)7ec: exit status [12](https://github.com/containerd/rust-extensions/actions/runs/4825275170/jobs/8595977056?pr=139#step:7:13)8:
	fatal: unable to access 'https://go.googlesource.com/mobile/': The requested URL returned error: 502
```

crates/shim/src/synchronous/mod.rs
jsturtevant: I added a little helper struct with an `Option` wrapped around it to get around the limitation for the signal handling function. (70934a20fff1eb6570b845f1c03fa8bf92caa601). I think it was useful for the signal functions.  There is probably an opportunity to improve signal_hook to support windows for that API longer term. I could do it for the listener too but I am not sure it helps with code readability in that case.  

Let me know what you think.

crates/shim/src/lib.rs
```diff
@@ -171,9 +182,18 @@ pub fn socket_address(socket_path: &str, namespace: &str, id: &str) -> String {
...
-    format!("unix://{}/{:x}.sock", SOCKET_ROOT, hash)
+    #[cfg(unix)]
+    {
+        format!("unix://{}/{:x}.sock", SOCKET_ROOT, hash)
+    }
+
+    #[cfg(windows)]
+    {
+        format!(r"\\.\pipe\containerd-shim-{}-pipe", hash)
+    }
```
mxpv, line 193: nit: if/else might be better here:

Like:
```rust
let machine_kind = if cfg!(unix) {
  "unix"
} else if cfg!(windows) {
  "windows"
} else {
  "unknown"
};
```

jsturtevant: and return the formatted strings in the if/else branches?

mxpv: Right, this will be a compile time if/else branches.

jsturtevant: Updated

jsturtevant:
I think this is ready for review, just waiting for a few merges and a release from ttrpc-rust which should be soon 🤞 .  I will leave temporary commits and leave it in draft until then.

jsturtevant:
nothing major changed in the ttrpc version I had pinned and the released crate.  Not sure if something else changed to cuase the integration tests to fail.  I might split out the ttrpc verion upgrade anyways (not sure what this repositories standards are but that is generally asked for in other repos I've worked in).

jsturtevant:
oh, looks like nightly CI started failing last week: https://github.com/containerd/rust-extensions/actions with the first failure in https://github.com/containerd/rust-extensions/actions/runs/5273246763

jsturtevant:
ttrpc 0.8.0 has been released!  I've moved this from draft to ready for review.

> I might split out the ttrpc verion upgrade anyways (not sure what this repositories standards are but that is generally asked for in other repos I've worked in).

@mxpv thoughts?

jsturtevant:
ci is passing again after #148

mxpv (approved):

PR #150, opened 2023-06-23
Pass all commandline arguments through

jsturtevant:
Because the second invocation of the shim doesn't have the containerd pipe passed to it, a shim that wants to communicate over the pipe needs to parse the arguments its own.  This makes it so the library pass all the arguments, which has already parsed the arguments allowing shims to use the containerd address.

This came up in runwasi: https://github.com/containerd/runwasi/pull/147#pullrequestreview-1495745320

### DIFF
--- a/crates/shim/src/synchronous/mod.rs
+++ b/crates/shim/src/synchronous/mod.rs
@@ -62,7 +62,8 @@ use util::{read_address, write_address};
 
 use crate::{
     api::DeleteResponse,
-    args, logger,
+    args::{self, Flags},
+    logger,
     protos::{
         protobuf::Message,
         shim::shim_ttrpc::{create_task, Task},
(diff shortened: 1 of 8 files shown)


crates/shim/src/synchronous/mod.rs
```diff
@@ -111,10 +112,9 @@ pub trait Shim {
     ///
     /// # Arguments
     /// - `runtime_id`: identifier of the container runtime.
-    /// - `id`: identifier of the shim/container, passed in from Containerd.
-    /// - `namespace`: namespace of the shim/container, passed in from Containerd.
+    /// - `args`: command line arguments passed to the shim which includes namespace and id
     /// - `config`: for the shim to pass back configuration information
-    fn new(runtime_id: &str, id: &str, namespace: &str, config: &mut Config) -> Self;
+    fn new(runtime_id: &str, args: &Flags, config: &mut Config) -> Self;
```
Burning1020, line 117: Could you please update `new()` in `crates/shim/src/asynchronous/mod.rs`

jsturtevant: I meant to do this, thanks for catching it!

mxpv (approved):

mxpv:
@jsturtevant could you pls resolve conflicts?

jsturtevant:
@mxpv updated, sorry for the delay was away from the computer for awhile!

Issue #134, opened 2023-04-12
Proposal: remove sync code


Burning1020:
In Rust, synchronous (sync) code is blocking, which means it will wait for a response before continuing with the next task. This can lead to slow performance if there are many synchronous operations.

On the other hand, asynchronous (async) code is non-blocking, which means it can perform other tasks while waiting for a response. This allows for better performance and can be especially useful in network applications where there are many concurrent requests.

Initially, I was hesitant about using async shim because it wasn't very stable in testing. Additionally, the Rust async ecosystem is still evolving and some libraries may not have full support for async yet. However, after doing some research, I found that most mainstream rust crates are now using async crates. If there are any bugs in async shim, we should work on fixing them instead of relying on sync code. 

Therefore, I decide to **stop developing sync code**. If you have any opinion, please leave a comment below. If there are no objections after one month(May 15th), we'll proceed with the plan to remove sync code in the next version.

Before we officially deprecate it, I will still commit PRs to fix some bugs.

/cc @mxpv @jiangliu @abel-von @ytoml @wedsonaf @AkihiroSuda

jsturtevant:
[Runwasi ](https://github.com/containerd/runwasi) is using the sync code today /cc @cpuguy83

cpuguy83:
"sync" code just needs to use threads instead of an async runtime.

I actually do not see how any shim would ever have enough traffic hitting it that using async would make any real difference here.

Burning1020:
Yes, maybe I haven't made it cleaer. We have many funcs written by both sync and async, they have the same function so that we need maintance two codes. Like [async fn notify_topic](https://github.com/containerd/rust-extensions/blob/main/crates/shim/src/asynchronous/monitor.rs#L112) and [fn notify_topic](https://github.com/containerd/rust-extensions/blob/main/crates/shim/src/synchronous/monitor.rs#L107)

What I want to do is to remove funcs both have sync and async, if you want do call async func in sync func, try tokio block_on? https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on

mxpv:
@Burning1020 we can start with removing sync code from the `runc-shim` crate to reduce maintenance burden. As for `shim` crate, I'd prefer to keep it slim and implementation agnostic, so we can move monitor stuff up the `runc-shim`, and continue to keep a small portion of sync/async code (like publisher).

PR #155, opened 2023-07-24
Remove sync code from runc shim

mxpv:
Closes: https://github.com/containerd/rust-extensions/issues/134


mxpv:
@jsturtevant any ideas why windows integration tests fail? Not sure I get what exactly wrong with it and I don't have windows env to verify (this also fails in `main`).

```
error[E0432]: unresolved import `ttrpc::asynchronous`
  --> crates\shim-protos\src\lib.rs:93:20
   |
93 |     pub use ttrpc::asynchronous::Client;
   |                    ^^^^^^^^^^^^ could not find `asynchronous` in `ttrpc`
```

jsturtevant:
looks like bringing in the `async` feature somehow. I will take a look today. 

> this also fails in main

Looks like main is failing due to https://github.com/containerd/rust-extensions/issues/160?

mxpv:
Let me rebase this and give it another try.

jsturtevant:
rebase didn't help 😞.  

CI on main is passing today: https://github.com/containerd/rust-extensions/actions/runs/5743079375 and I was able to run the integration test locally on main. Going to pull this PR to see I can track down what is causing the mix up

jsturtevant:
Still tracking down what is happening but `cargo build -p containerd-shim-protos` and `cargo build -p containerd-shim` builds fine on windows.

but `cargo build --example skeleton` fails with message `error: could not compile containerd-shim-protos (lib) due to 117 previous error`

jsturtevant:
@mxpv I don't know why but it seems the examples are getting the `feature=async` enabled on them after the changes in this PR.  Maybe the changes in the `runc crate` are causing the compiler/cargo to pick that up by default?

If I run `cargo build --example skeleton -p containerd-shim` then this works.  So the fix would be to add `-p containerd-shim` to the cargo calls in:

https://github.com/containerd/rust-extensions/blob/27a503a474c68cdba7eff5b9f1b439d1846c39e0/.github/workflows/ci-windows.yml#L97-L99

mxpv:
@jsturtevant That did the trick!

jsturtevant:
LGTM

Issue #122, opened 2023-03-12
Fix integration tests


mxpv:
Follow up for https://github.com/containerd/rust-extensions/issues/110
There are a couple of integration tests that we disabled on CI (https://github.com/containerd/rust-extensions/pull/104) which require investigation:
- `TestContainerPTY`
- `TestContainerExecLargeOutputWithTTY`
- `TestTaskUpdate`
- `TestTaskResize`

We should fix the shim and enable them back.

jsturtevant:
I was able to reproduce the first three but the last one works:

```
sudo TEST_RUNTIME="io.containerd.runc.v2-rs" EXTRA_TESTFLAGS="-run TestTaskResize" make integration

INFO[0000] running tests against containerd              
revision=3dce8eb055cbb6872793272b4f20ed16117344f8 runtime=io.containerd.runc.v2-rs snapshotter= version=1.6.21
INFO[0000] start to pull seed image                      image="ghcr.io/containerd/busybox:1.32"
=== RUN   TestTaskResize
=== PAUSE TestTaskResize
=== CONT  TestTaskResize
--- PASS: TestTaskResize (1.17s)

ok      github.com/containerd/containerd/integration/client     3.028s
```

jsturtevant:
For TaskRisize I have a script that ran it continously and it is flaky:

 on the 190th run locally:
```

INFO[0000] start to pull seed image                      image="ghcr.io/containerd/busybox:1.32"
=== RUN   TestTaskResize
=== PAUSE TestTaskResize
=== CONT  TestTaskResize
    container_test.go:2357: failed to create shim task: ttrpc: closed: unknown
containerd logs confirm it was during create:

time="2023-08-11T16:42:23.441614347Z" level=debug msg="event published" ns=testing topic=/snapshot/prepare type=containerd.events.SnapshotPrepare
time="2023-08-11T16:42:23.447312550Z" level=debug msg="get snapshot mounts" key=TestTaskResize
time="2023-08-11T16:42:23.460445957Z" level=debug msg="event published" ns=testing topic=/containers/create type=containerd.events.ContainerCreate
time="2023-08-11T16:42:23.486143669Z" level=debug msg="get snapshot mounts" key=TestTaskResize
time="2023-08-11T16:42:23.901091473Z" level=debug msg="garbage collected" d=7.965004ms
time="2023-08-11T16:42:27.508985837Z" level=info msg="shim disconnected" id=TestTaskResize
```

PR #166, opened 2023-08-03
Use Go version associated with containerd release in CI

### LINKED ISSUE #147: Nightly CI is failing integration tests
Looks like nightly CI started failing last week: https://github.com/containerd/rust-extensions/actions with the first failure in https://github.com/containerd/rust-extensions/actions/runs/5273246763

_Originally posted by @jsturtevant in https://github.com/containerd/rust-extensions/issues/139#issuecomment-1601821216_

Interestingly it was passing and is now failing on the same commit: [35a97f1](https://github.com/containerd/rust-extensions/commit/35a97f17d55753bb1ef04c28cd7c3203993932b0) so something else in the dependency chain must have changed.

jsturtevant:
This ensures that there isn't failures with testing when go version doesn't match the release as experienced in #147.

this also bumps the containerd versions to latest patches

fixes #147

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,16 +100,22 @@ jobs:
       - name: Checkout extensions
         uses: actions/checkout@v3
 
-      - uses: actions/setup-go@v4
-        with:
-          go-version: '1.20.4'
-
       - name: Checkout containerd
         uses: actions/checkout@v3
         with:
           repository: containerd/containerd
           path: src/github.com/containerd/containerd
           ref: ${{ matrix.containerd }}
+
+      - name: Get Go Version
+        run: |
+          go_version=$(awk -F': ' '/GO_VERSION/ {gsub(/["'\'']/, "", $2); print $2; exit}' .github/workflows/release.yml)
+          echo "GO_VERSION=$go_version" >> $GITHUB_ENV
+        working-directory: src/github.com/containerd/containerd
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
 
       - name: Install containerd
         env:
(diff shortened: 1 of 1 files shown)


jsturtevant:
When looking into this I found that since we need to build the tests from source anyways, we could just use the go version.

Got it working but some tests are skipped (go test -test.skip) in CI (https://github.com/containerd/rust-extensions/issues/122) using a flag that is only supported in golang 1.20.  

Could download the binaries and just use go 1.20 for the test suite or look into the test failures.

Mossaka (approved):
LGTM!

jsturtevant:
unfortunately CI isn't passing do the skipped tests so will need to fix those or only use golang 1.20+ for the tests

jsturtevant:
@Mossaka @mxpv this is ready for review, It also includes the temp fix for CI failing. Alternatively I also put the temp fix in https://github.com/containerd/rust-extensions/pull/179

.github/workflows/ci.yml
```diff
@@ -100,17 +100,23 @@ jobs:
...
       - name: Checkout containerd
         uses: actions/checkout@v3
         with:
           repository: containerd/containerd
           path: src/github.com/containerd/containerd
           ref: ${{ matrix.containerd }}
 
+      - name: Get Go Version
+        run: |
+          go_version=$(awk -F': ' '/GO_VERSION/ {gsub(/["'\'']/, "", $2); print $2; exit}' .github/workflows/release.yml)
```
mxpv, line 112: We don't have `.github/workflows/release.yml`, right?

```
❯ cat .github/workflows/release.yml
cat: .github/workflows/release.yml: No such file or directory
```

jsturtevant: this is the containerd release file since it is the containerd repository we checked out in the step before

mxpv: oh, right right.

mxpv:
Could you also squash the commit history?

mxpv (approved):

PR #167, opened 2023-08-03
feat(runc-shim): add a version flag

### LINKED ISSUE #120: Proposal: Add `--version` flag to print the build info of the binary
It would be nice to have a --version arg for the shims to be able to determine the build version information.

Mossaka:
This PR adds two things
1. expose the `parse` function from `containerd-shim` crate and move the namespace check to the boostrap function
2. implement git hash for helping the version information on the runc-shim. 

Closes #120

### DIFF
--- a/crates/runc-shim/build.rs
+++ b/crates/runc-shim/build.rs
@@ -1,0 +1,39 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+use std::{process::Command, str::from_utf8};
+
+fn main() {
+    let output = match Command::new("git").arg("rev-parse").arg("HEAD").output() {
+        Ok(output) => output,
+        Err(_) => {
+            return;
+        }
+    };
+    let mut hash = from_utf8(&output.stdout).unwrap().trim().to_string();
+
+    let output_dirty = match Command::new("git").arg("diff").arg("--exit-code").output() {
+        Ok(output) => output,
+        Err(_) => {
+            return;
+        }
+    };
+
+    if !output_dirty.status.success() {
+        hash.push_str(".m");
+    }
+    println!("cargo:rustc-env=CARGO_GIT_HASH={}", hash);
+}
--- a/crates/runc-shim/src/main.rs
+++ b/crates/runc-shim/src/main.rs
@@ -14,7 +14,9 @@
    limitations under the License.
 */
 
-use containerd_shim::asynchronous::run;
+use std::env;
+
+use containerd_shim::{asynchronous::run, parse};
 
 mod common;
 mod console;
@@ -27,7 +29,21 @@ mod task;
 
 use service::Service;
 
+fn parse_version() {
+    let os_args: Vec<_> = env::args_os().collect();
+    let flags = parse(&os_args[1..]).unwrap();
+    if flags.version {
+        println!("{}:", os_args[0].to_string_lossy());
+        println!("  Version: {}", env!("CARGO_PKG_VERSION"));
+        println!("  Revision: {}", env!("CARGO_GIT_HASH"));
+        println!();
+
+        std::process::exit(0);
+    }
+}
+
 #[tokio::main]
 async fn main() {
+    parse_version();
     run::<Service>("io.containerd.runc.v2-rs", None).await;
 }
--- a/crates/shim/src/args.rs
+++ b/crates/shim/src/args.rs
@@ -39,15 +39,17 @@ pub struct Flags {
     /// Shim action (start / delete).
     /// See <https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L191>
     pub action: String,
+    /// Version of the shim.
+    pub version: bool,
 }
 
 /// Parses command line arguments passed to the shim.
-/// This func replicates https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L110
 pub fn parse<S: AsRef<OsStr>>(args: &[S]) -> Result<Flags> {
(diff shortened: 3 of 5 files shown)


crates/runc-shim/build.rs
```diff
@@ -0,0 +1,25 @@
+use std::process::Command;
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
```
mxpv, line 4: I believe cargo will always run `build.rs` no matter what.

mxpv (approved):

Burning1020:
OK, I'll review it ASAP

crates/runc-shim/src/main.rs
```diff
@@ -27,7 +29,21 @@ mod task;
...
+        println!("  Revision: {}", env!("CARGO_GIT_HASH"));
+        println!();
+
+        std::process::exit(0);
+    }
+}
+
 #[tokio::main]
 async fn main() {
+    parse_version();
```
Burning1020, line 47: Why not put it to `bootstrap`, if not, flags will be parsed twice.
And this version flag should be a common code in shim, not in runc-shim

crates/runc-shim/src/main.rs
```diff
@@ -27,7 +29,21 @@ mod task;
 
 use service::Service;
 
+fn parse_version() {
+    let os_args: Vec<_> = env::args_os().collect();
+    let flags = parse(&os_args[1..]).unwrap();
+    if flags.version {
+        println!("{}:", os_args[0].to_string_lossy());
+        println!("  Version: {}", env!("CARGO_PKG_VERSION"));
+        println!("  Revision: {}", env!("CARGO_GIT_HASH"));
```
Burning1020, line 38: Is it possiable to add "dirty" if some codes changed but witout committed?

Burning1020 (changes requested):
Could you please rebase your commits? Because there is a Merge commit.

crates/runc-shim/src/main.rs
Mossaka: I am trying to understand the purpose of this info - is it mostly for local debugging?

crates/shim/src/args.rs
```diff
@@ -39,15 +39,18 @@ pub struct Flags {
     /// Shim action (start / delete).
     /// See <https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L191>
     pub action: String,
+    /// Version of the shim.
+    pub version: bool,
 }
 
 /// Parses command line arguments passed to the shim.
-/// This func replicates https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L110
+/// This func replicates <https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L110>
```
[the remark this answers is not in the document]
Mossaka: Sure

crates/runc-shim/src/main.rs
jsturtevant: yes, I think it is mostly useful when running locally, if you think you are on the wrong commit but actually have a few commits in it. 

```
git describe --tags --dirty --long  
```

 should give you that information

Mossaka:
```bash
containerd-shim-runc-v2:
  Version:  v1.7.3
  Revision: 7880925980b188f4c97b462f709d0db8e8962aff
  Go version: go1.20.6
```

This is what runc produces for version. How closely do we want to follow it? Obviously the Go version doesn't make sense here...

Mossaka: And this is what this PR adds

```bash
./target/debug/containerd-shim-runc-v2-rs:
  Version: 0.1.1
  Revision: 60dc92b9d56a2fb5fc9b58e02b1e7929320608c0

Comment (Burning1020) [reply crates/runc-shim/src/main.rs:38]:
```bash
$ containerd --version
containerd github.com/containerd/containerd e86e562da.m e86e562dad79a422b6b5a8ec2292ffc61852cf09.m
```
containerd binary already has this abaility, printing".m" if the repo is changed and uncommitted.
In another word, if the output of `git status -s` is empty, it means that all changes are committed. If not, a "dirty" or ".m" should be append to git revision.

I can do that lately if that is not convenience for you.

Mossaka: I can add `.m` to the postfix of revision

Mossaka: Done.

Burning1020: OK. Thanks

crates/runc-shim/src/main.rs
Burning1020: I still prefer to put it to `bootstrap` for the clean code principle. This code can be uesed by other shim too. We can add another custom env to store the runc-shim pkg version and then print it in shim.

crates/runc-shim/build.rs
```diff
@@ -0,0 +1,23 @@
+use std::{process::Command, str::from_utf8};
```
Burning1020, line 1: Need Copyright declaration in the first line. @mxpv Did we miss some file header check in CI?

crates/runc-shim/src/main.rs
Mossaka: >  We can add another custom env to store the runc-shim pkg version and then print it in shim.

Could you please elaborate on this point? I just can't figure out how the env variables built in `runc-shim` could be used in `shim` crate. `runc-shim` is depending on `shim`, so rustc will build `shim` first, and then `runc-shim` second. When building `shim` crate, rustc will not be able to see the env variables that `runc-shim` adds. Am I missing something here?

crates/runc-shim/build.rs
Mossaka: Done.

jsturtevant:
you've hit https://github.com/containerd/rust-extensions/issues/147#issuecomment-1684250819 with the integration tests.

crates/runc-shim/src/main.rs
Burning1020: I got you. You are right. If we only provide git hash id, we can move all the parse_version in `shim`, but with pkg version, that's pretty hard!

Burning1020 (approved):

Burning1020 (commented):
LGTM

jsturtevant:
@Mossaka a rebase should fix the integration test failing

Mossaka:
Let me try rerunning the CI

Mossaka:
Once we get this PR merged, can we get a new release cut?

Mossaka:
@mxpv could we merge this PR?

PR #171, opened 2023-08-09
Improve containerd client ergonomics

mxpv:
Make containerd client a bit more user friendly.

Instead of

```rust
   let channel = client::connect("/run/containerd/containerd.sock")
        .await
        .expect("Connect Failed");

    let mut client = VersionClient::new(channel);
    let resp = client.version(()).await.expect("Failed to query version");
```

we can do

```rust
    let client = Client::from_path("/var/run/containerd/containerd.sock")
        .await
        .expect("Connect failed");

    let version_client = client.version()
    let task_client = client.tasks()
    ...
```

### DIFF
--- a/crates/client/examples/version.rs
+++ b/crates/client/examples/version.rs
@@ -14,18 +14,20 @@
    limitations under the License.
 */
 
-use client::services::v1::version_client::VersionClient;
-use containerd_client as client;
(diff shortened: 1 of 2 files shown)


Mossaka (approved):
LGTM!

PR #177, opened 2023-08-10
Handle cgroup v1 or v2 to enable resource updates

jsturtevant:
Part of #122 

This test was failing in 22.04  because it has cgroups v2.  This follows same pattern as the metrics to select either v1 or v2.

### DIFF
--- a/crates/shim/src/cgroup.rs
+++ b/crates/shim/src/cgroup.rs
@@ -97,26 +97,7 @@ fn write_process_oom_score(pid: u32, score: i64) -> Result<()> {
 pub fn collect_metrics(pid: u32) -> Result<Metrics> {
     let mut metrics = Metrics::new();
 
-    let hierarchies = hierarchies::auto();
-    let cgroup = if hierarchies.v2() {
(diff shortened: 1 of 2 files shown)


crates/shim/src/cgroup.rs
```diff
@@ -193,12 +174,35 @@ pub fn collect_metrics(pid: u32) -> Result<Metrics> {
     Ok(metrics)
 }
 
+// get_cgroup will return either cgroup v1 or v2 depending on
```
dcantah, line 177: Comment got cut off

jsturtevant: :facepalm:

jsturtevant:
same go get error again (but other integration tests passed just fine :-/)

jsturtevant:
go get is stuggling today, I will try again Monday morning.

dcantah (approved):

jsturtevant:
Same go get issue today.  I will need to look deeper :-/

mxpv (approved):

PR #190, opened 2023-09-05
Fix CI

mxpv:
Dependabot opened a bunch of PRs, which are failing because of breaking change in ttrpc 0.8.1 (https://github.com/containerd/ttrpc-rust/pull/197).
This PR pins ttrpc crate to 0.8.0 to make CI green until we figure out the fix.


jsturtevant:
LGTM

Issue #197, opened 2023-09-15
Proposal: Support content fetch


bryantbiggs:
It would be nice if fetching content was supported to be able to pull images via the client https://github.com/containerd/containerd/blob/main/cmd/ctr/commands/content/fetch.go

PR #181, opened 2023-08-24
Remove temporary work around and bump containerd versions

jsturtevant:
[Containerd 1.7.4](https://github.com/containerd/containerd/releases/tag/v1.7.4) has the fix for the dependency issue fixed previously in https://github.com/containerd/rust-extensions/pull/179

PR #198, opened 2023-09-16
feat: Update proto files to latest to add `transfer` and `streaming` services; remove gogo use

bryantbiggs:
- Updates proto files with latest from containerd mainline
	- This removes gogo use per https://github.com/containerd/containerd/pull/6867
	- Adds `streaming` and `transfer` protos
- Update the versions of containerd used in CI to latest since the protos are being updated

May relate to #197 (`transfer` service)


mxpv:
Thanks for the contribution! I have a few suggestions:
- New proto services must be registered in the [`build.rs`](https://github.com/containerd/rust-extensions/blob/9388ecf69994e4d019df007ac35580f86690dd56/crates/client/build.rs#L19) to generate the actual Rust code.
- We don't use `runtime` services in `containerd-client`, since it's a client to the containerd daemon, not runtimes the daemon executes. So those can be removed.

mxpv (changes requested):
Thanks for the contribution! I have a few suggestions:

- New proto services must be registered in the [build.rs](https://github.com/containerd/rust-extensions/blob/9388ecf69994e4d019df007ac35580f86690dd56/crates/client/build.rs#L19) to generate the actual Rust code.
- We don't use runtime services in containerd-client, since it's a client to the containerd daemon, not runtimes the daemon executes. So those can be removed.

bryantbiggs:
Not sure if I should do anything further to the CI or wait for something like  #181

jsturtevant:
> Not sure if I should do anything further to the CI or wait for something like #181

the version bump for CI is required?  I haven't been able to track down why bumping the containerd minor versions is causing a failure.

mxpv:
Let's leave CI changes out and do it in an a separate PR?

bryantbiggs:
> Let's leave CI changes out and do it in an a separate PR?

Sounds good! I don't know if there is a min supported version for the proto changes here, which is why I just bumped them out of caution. I'll remove those now

mxpv (approved):

PR #201, opened 2023-09-19
deps: Remove the use of the `regex` crate

bryantbiggs:
- Instead of using the `regex` crate to ensure the path ends at the correct `/` which is a bit heavy handed for this use case, it has been replaced with an iterator based implementation 
- Simplify longest common prefix logic -  this was pinched from [here](https://users.rust-lang.org/t/is-this-code-idiomatic/51798/4), the original logic seemed a bit verbose

### DIFF
--- a/crates/shim/src/mount.rs
+++ b/crates/shim/src/mount.rs
@@ -29,7 +29,6 @@ use log::error;
 use nix::mount::{mount, MsFlags};
 #[cfg(target_os = "linux")]
 use nix::unistd::{fork, ForkResult};
-use regex_lite::Regex;
 
 use crate::error::{Error, Result};
 #[cfg(not(feature = "async"))]
@@ -249,14 +248,13 @@ fn longest_common_prefix(dirs: &[String]) -> &str {
     for (i, byte) in first_dir.as_bytes().iter().enumerate() {
         for dir in dirs {
             if dir.as_bytes().get(i) != Some(byte) {
-
                 let mut end = i;
                 // guaranteed not to underflow since is_char_boundary(0) is always true
                 while !first_dir.is_char_boundary(end) {
                     end -= 1;
                 }
 
-                return &first_dir[0..end]
+                return &first_dir[0..end];
             }
         }
     }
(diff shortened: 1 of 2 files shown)


mxpv:
I wonder if we need regex dependency at all.
we only use it in `trim_flawed_dir`.

Given these test cases:
```
tcases.push(("/.foo-_bar/foo", "/.foo-_bar/".to_string()));
tcases.push(("/.foo-_bar/foo/", "/.foo-_bar/foo/".to_string()));
tcases.push(("/.foo-_bar/foo/bar", "/.foo-_bar/foo/".to_string()));
tcases.push(("/.foo-_bar/foo/bar/", "/.foo-_bar/foo/bar/".to_string()));
```

I feel like it should be relatively straightforward to rewrite it without regular expressions.

WDYT?

bryantbiggs:
> I wonder if we need regex dependency at all. we only use it in `trim_flawed_dir`.
> 
> Given these test cases:
> 
> ```
> tcases.push(("/.foo-_bar/foo", "/.foo-_bar/".to_string()));
> tcases.push(("/.foo-_bar/foo/", "/.foo-_bar/foo/".to_string()));
> tcases.push(("/.foo-_bar/foo/bar", "/.foo-_bar/foo/".to_string()));
> tcases.push(("/.foo-_bar/foo/bar/", "/.foo-_bar/foo/bar/".to_string()));
> ```
> 
> I feel like it should be relatively straightforward to rewrite it without regular expressions.
> 
> WDYT?

Great point! Let me know what you think of the implementation

crates/shim/src/mount.rs
```diff
@@ -239,49 +238,52 @@ fn options_size(options: &[String]) -> usize {
...
 fn trim_flawed_dir(s: &str) -> String {
-    let r = Regex::new(r"((/[^/]+)+/)([^/]*)").unwrap();
-    r.replace(s, "$1").to_string()
+    match s.ends_with('/') {
+        true => s.to_owned(),
+        false => {
+            // Iterate in reverse order to find the last '/', then return string in correct order
+            s.chars()
+                .rev()
+                .skip_while(|x| *x != '/')
```
mxpv, line 276: Would [`rfind`](https://doc.rust-lang.org/std/string/struct.String.html#method.rfind) be a better fit for this?

bryantbiggs: #TIL - yes, that looks much better, thank you! Updated in https://github.com/containerd/rust-extensions/pull/201/commits/9e2c7a7113a1858b8e273fe94d1abd35a09b8ea9

Burning1020 (approved):
Nice

PR #202, opened 2023-09-20
feat: Add helper script to aid in keeping proto files up to date

bryantbiggs:
- Add helper script to aid in keeping proto files up to date
	- Its not perfect but it helps with the bulk of the task - there will still be additional, manual involvement to update/resolve any changes in the rust code
	- It currently is only updating protos from the `containerd` organization, but can be extended to the other vendored sources if this seems valuable/useful
- There are a few changes related to `cgroups`:
	1. Both `v1` and `v2` are being pulled in so they have been namespaced under their respective version. Since the existing implementation is using `v1`, that has been maintained through this change. In a later PR, we can look at if/how to expose `v2` to users behind a feature flag
	2. I suspect I am doing something wrong or mis-understanding because where is `github.com/containerd/cgroups/v3` ?! I see it used here and referenced across the `contaienrd` project but I don't see where that `v3` is located or coming from. If I go to the `cgroups` repository I only see `cgroup1/`, `cgroup2/` and `cmd/` directories
- Upstream PR to fix a missing namespace that causes tonic to crash https://github.com/containerd/containerd/pull/9126
	
Open to thoughts/feedback - and I've loved to help document some of this once its ironed out


crates/runc-shim/src/runc.rs
```diff
@@ -350,11 +350,6 @@ impl ProcessLifecycle<InitProcess> for RuncInitLifecycle {
 
 impl RuncInitLifecycle {
     pub fn new(runtime: Runc, opts: Options, bundle: &str) -> Self {
-        let work_dir = Path::new(bundle).join("work");
```
bryantbiggs, line 353: is this kosher since `criu_path` is no longer in the proto spec?

jsturtevant:
> 2\. I suspect I am doing something wrong or mis-understanding because where is `github.com/containerd/cgroups/v3` ?! I see it used here and referenced across the `contaienrd` project but I don't see where that `v3` is located or coming from. If I go to the `cgroups` repository I only see `cgroups1/`, `cgroups2/` and `cmd/` directories

This took me a while to figure out too, It is coming from https://github.com/containerd/cgroups/blob/v3.0.2/go.mod#L1

crates/shim-protos/vendor/github.com/containerd/cgroups/cgroup1/stats/metrics.proto
jsturtevant, line 1: I think we would want to keep the v3/cgroup1 designation, to indicate the is version 3 of the cgroup?

bryantbiggs: hmm, yes I think you are correct. However, it does raise a few questions:
1. If we are adding this version into the vendored path - should we do the same for the other vendored protos? At minimum I will add some documentation to state where the v3 is coming from - but it would be nice to make this consistent across the vendored protos if possible
2. As a consumer of the libraries here - how would we handle different project versions? For example - if cgroups released a v4.0, how would that be reflected or handled in this project?

scripts/update-protos.sh
```diff
@@ -0,0 +1,93 @@
...
+  done
+}
+
+### Update the `crates/shim-protos` `containerd` proto files
+function update_shim_protos_containerd {
+  local CRATE_DIR="${DIR}/crates/shim-protos/vendor/github.com/containerd/containerd"
+
+  # Copy api/ proto files
+  rm -rf ${CRATE_DIR}/api
+  find ${CONTAINERD_DIR}/api -type f -name \*.proto -regextype posix-egrep -regex '.*\api/(events|runtime|services/ttrpc|types).*' ! -path '*/transfer/*' -print0 | while IFS= read -r -d $'\0' file;
```
mxpv, line 45: Could you pls add a human readable comment, whats added/included/filtered by regex, etc

scripts/update-protos.sh
```diff
@@ -0,0 +1,93 @@
...
+#
+# Note: This currently just copies the proto files from the upstream containerd projects
+# into their respective vendored directories. There may be additional work involved to
+# update the respective crates (manually) after running the script.
+
+DIR=$(dirname "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )")
+TEMP_DIR=`mktemp -d -p "$DIR"`
+
+CONTAINERD_DIR="${TEMP_DIR}/containerd"
+git clone https://github.com/containerd/containerd --depth 1 $CONTAINERD_DIR
```
mxpv, line 13: Could you pls add `git clone --branch $CONTAINERD_BRANCH`, so we can control which protos we're getting here.

> -b <name>, --branch <name>
> Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to <name> branch instead. In a non-bare repository, this is the branch that will be checked out.  --branch can also take tags and detaches the HEAD at that commit in the resulting repository.

crates/shim-protos/vendor/github.com/containerd/cgroups/cgroup1/stats/metrics.proto
jsturtevant: I guess there would be two approaches:
 - have support for v3 and v4 at the same time in the library here and eventually removing older versions 
 - have support for one version in the library and to upgrade you would need to move the new version of the library to get the new protobuffs

The second is probably the easier to maintain?

bryantbiggs:
I need to revisit this when I get some time - it doesn't work well across platform (Linux vs MacOS) as its written

mxpv:
This was solved in https://github.com/containerd/rust-extensions/pull/244, so I'm going to close this one. Let us know if there any missing bits left.

PR #200, opened 2023-09-19
chore: Standardize Cargo.toml files, enable more workspace dependencies

bryantbiggs:
- Attempt to unify and standardize the Cargo.toml files used through the project
	- Order the dependencies alphabetically; the async deps have been kept as a separate section but happy to include this in the mix as well
	- In the future it would be nice to use `cargo-sort` and have a CI check to keep it consistent but currently it [lacks support for workspace inheritence](https://github.com/DevinR528/cargo-sort/issues/54)
	- Hoist up dependencies to the main Cargo.toml and use the workspace dependency inheritance to align versions
	- Add `keywords` and `categories` sections where missing
-  Add the client examples to its Cargo.toml

PR #203, opened 2023-09-21
ci: Download containerd binaries from GitHub release archives for integration tests

bryantbiggs:
- Download containerd binaries from GitHub release archives for integration tests

When investigating the merge queue integration failures on https://github.com/containerd/rust-extensions/pull/200 I came across https://github.com/containerd/rust-extensions/issues/147 and the discussion around using the release archives instead of building in the pipeline. I created a quick fork in a separate org and pushed up some changes to validate since this won't run the changes on my PR here, it will only use `main` (its Github security feature). So here is the results of these changes https://github.com/clowdhaus/rust-extensions/actions/runs/6267161577

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,43 +89,29 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04]
-        containerd: [v1.6.21, v1.7.1]
+        containerd: [1.6.21, 1.7.1]
 
     steps:
       - name: Checkout extensions
         uses: actions/checkout@v4
+
+      - name: Download containerd release v${{ matrix.containerd }} archive
+        uses: robinraju/release-downloader@v1.8
+        with:
+          repository: "containerd/containerd"
+          tag: v${{ matrix.containerd }}
+          fileName: "containerd-${{ matrix.containerd }}-linux-amd64.tar.gz"
+          extract: true
+
+      - name: Add containerd binaries to PATH
+        run: echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
 
       - name: Checkout containerd
         uses: actions/checkout@v4
         with:
           repository: containerd/containerd
           path: src/github.com/containerd/containerd
-          ref: ${{ matrix.containerd }}
-
-      - name: Get Go Version
-        run: |
-          go_version=$(awk -F': ' '/GO_VERSION/ {gsub(/["'\'']/, "", $2); print $2; exit}' .github/workflows/release.yml)
-          echo "GO_VERSION=$go_version" >> $GITHUB_ENV
-        working-directory: src/github.com/containerd/containerd
-
-      - uses: actions/setup-go@v4
-        with:
-          go-version: ${{ env.GO_VERSION }}
-
-      - name: Install containerd
-        env:
-          GOFLAGS: -modcacherw
-          CGO_ENABLED: 1
-        run: |
-          # Install containerd dependencies first
-          sudo apt-get install -y gperf
-          sudo -E PATH=$PATH script/setup/install-seccomp
-          sudo -E PATH=$PATH script/setup/install-runc
-          sudo -E PATH=$PATH script/setup/install-cni $(grep containernetworking/plugins go.mod | awk '{print $2}')
-          # Install containerd
-          sudo -E PATH=$PATH make bin/containerd GO_BUILD_FLAGS="-mod=vendor" BUILDTAGS="no_btrfs no_devmapper"
-          sudo -E PATH=$PATH install bin/containerd /usr/local/bin/
-        working-directory: src/github.com/containerd/containerd
+          ref: v${{ matrix.containerd }}
 
       - name: Install shim
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -89,43 +89,29 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04]
-        containerd: [v1.6.21, v1.7.1]
+        containerd: [1.6.21, 1.7.1]
```
bryantbiggs, line 92: The release archive name does not have the preceding `v` so I removed it from the matrix and added it in the appropriate places below

.github/workflows/ci.yml
```diff
@@ -89,43 +89,29 @@ jobs:
...
       - name: Checkout extensions
         uses: actions/checkout@v4
 
+      - name: Download containerd release v${{ matrix.containerd }} archive
+        uses: robinraju/release-downloader@v1.8
+        with:
+          repository: "containerd/containerd"
+          tag: v${{ matrix.containerd }}
+          fileName: "containerd-${{ matrix.containerd }}-linux-amd64.tar.gz"
+          extract: true
```
mxpv, line 104: I think same can be achieved with gh CLI:

```
❯ gh release download --repo containerd/containerd --pattern '*linux-amd64.tar.gz' --output containerd.zip v1.7.3
```

.github/workflows/ci.yml
```diff
@@ -89,43 +89,33 @@ jobs:
...
-        containerd: [v1.6.21, v1.7.1]
+        containerd: [1.6.21, 1.7.1]
 
     steps:
       - name: Checkout extensions
         uses: actions/checkout@v4
 
+      - name: Download containerd archive
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
bryantbiggs, line 100: FYI: passing the default token is required when using in the workflow https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows

.github/workflows/ci.yml
bryantbiggs: yes! Updated in https://github.com/containerd/rust-extensions/pull/203/commits/ee2a2330f5b3867b068c7194102c8f4d190bef09

.github/workflows/ci.yml
```diff
@@ -89,43 +89,33 @@ jobs:
...
+        run: |
+          gh release download v${{ matrix.containerd }} \
+            --repo containerd/containerd \
+            --pattern 'containerd-${{ matrix.containerd }}-linux-amd64.tar.gz' \
+            --output containerd.tar.gz
+
+      - name: Extract containerd binaries to $HOME/.local/bin
+        run: |
+          mkdir -p $HOME/.local/bin
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
```
mxpv, line 110: why not to `/usr/local/bin/` ? it's already in `PATH`

.github/workflows/ci.yml
mxpv: Hmm.. 
I think archive name can be simplified to something like `--pattern '*-linux-amd64.tar.gz'` (as release version is provided separately via `gh release download ${{ matrix.containerd }} ... `).

.github/workflows/ci.yml
bryantbiggs: I tried to originally but it looks like GitHub has some restrictions on this 
image

.github/workflows/ci.yml
```diff
@@ -89,43 +89,33 @@ jobs:
...
-      - name: Install containerd
-        env:
-          GOFLAGS: -modcacherw
-          CGO_ENABLED: 1
-        run: |
-          # Install containerd dependencies first
-          sudo apt-get install -y gperf
-          sudo -E PATH=$PATH script/setup/install-seccomp
-          sudo -E PATH=$PATH script/setup/install-runc
-          sudo -E PATH=$PATH script/setup/install-cni $(grep containernetworking/plugins go.mod | awk '{print $2}')
```
jsturtevant, line 124: nice, I ended up not doing this since I thought cni and other parts where needed but looks like it isn't!

.github/workflows/ci.yml
bryantbiggs: I tried several different variations and no matter what it said there was unable to download since there is multiple matches. Setting it to the full archive name was the only one that worked
image

bryantbiggs: actually, I found one that works. let me update `gh release download v1.7.3 --repo containerd/containerd --pattern 'containerd-*-linux-amd64.tar.gz' --output containerd.tar.gz`

bryantbiggs: ok updated in https://github.com/containerd/rust-extensions/pull/203/commits/395f7374df3a3ad1d226aada932e38348f56fae8

jsturtevant:
LGTM

The failure looks like a temporary gitlab issue: `fatal: unable to access 'https://gitlab.com/cznic/sqlite/': The requested URL returned error: 502`

mxpv (approved):

PR #204, opened 2023-09-23
deps: Update `nix` crate to latest

bryantbiggs:
- Update `nix` crate to latest
	- One of the major changes between [`0.26` -> `0.27`](https://github.com/nix-rust/nix/blob/master/CHANGELOG.md#changed) is that all features are turned off by default and users need to opt into the features they wish to use

Closes #184


crates/runc-shim/src/common.rs
```diff
@@ -206,7 +203,6 @@ pub fn receive_socket(stream_fd: RawFd) -> containerd_shim::Result<RawFd> {
         "copy_console: console socket get path: {}, fd: {}",
         path, &fds[0]
     );
-    tcgetattr(fds[0])?;
```
bryantbiggs, line 209: after updating, this was throwing the following error:

```sh
error[E0277]: the trait bound `i32: std::os::fd::AsFd` is not satisfied
    --> crates/runc-shim/src/common.rs:209:15
     |
209  |     tcgetattr(fds[0])?;
     |     --------- ^^^^^^ the trait `std::os::fd::AsFd` is not implemented for `i32`
     |     |
     |     required by a bound introduced by this call
     |
```

Looking into the docs, it looks like the current implementation is a no-op since `tcsetattr()` is never called on the structure https://github.com/nix-rust/nix/blob/996db47d542ae20f09eb344b9fcb88c40ae38e3d/src/sys/termios.rs#L1156-L1158

Issue #207, opened 2023-09-27
ContainerAttach integration tests flaking


jsturtevant:
```
2023-09-27T16:54:46.0829155Z === CONT  TestContainerAttachProcess
2023-09-27T16:54:46.2301858Z     container_linux_test.go:716: write /proc/self/fd/29: broken pipe
2023-09-27T16:54:46.2314358Z     container_linux_test.go:724: write /proc/self/fd/29: broken pipe
2023-09-27T16:54:46.2384256Z     container_linux_test.go:746: expected output "hello\nhello\n" but received ""
2023-09-27T16:54:46.2591524Z --- FAIL: TestContainerAttachProcess (0.18s)
2023-09-27T16:54:46.2592190Z === CONT  TestContainerAttach
2023-09-27T16:54:46.3625257Z     container_linux_test.go:515: write /proc/self/fd/29: broken pipe
2023-09-27T16:54:46.3637557Z     container_linux_test.go:528: write /proc/self/fd/29: broken pipe
2023-09-27T16:54:46.3876891Z     container_linux_test.go:549: expected output "hello\nhello\n" but received ""
2023-09-27T16:54:46.3910662Z --- FAIL: TestContainerAttach (0.13s)
```

Occurrences: 
https://github.com/containerd/rust-extensions/actions/runs/6328911110/job/17188102327?pr=206
https://github.com/containerd/rust-extensions/pull/181#issuecomment-1736306658

jsturtevant:
https://github.com/containerd/rust-extensions/actions/runs/6329201731/job/17192981485

PR #208, opened 2023-09-28
Simplify doc management and update READMEs

mxpv:
Currently we manage root documentation for each crate in 2 places: crate's README.md on Github and module's root documentation (comments that start with `//!`) that gets uploaded to docs.rs. Typically this is the same text that needs to be present in 2 places.

This PR aims to simplify this flow and uses README.md in both places, so there is no need to sync docs. When building, each crate will use `include_str!="README.dm"` macro to include the latest docs from the file.

I avoid extra work on local environments, this feature is hidden behind `docs` toggle, which is automatically enabled when building documentation for docs.rs.

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -14,11 +14,10 @@
    limitations under the License.
 */
 
+#![cfg_attr(feature = "docs", doc = include_str!("../README.md"))]
 // No way to derive Eq with tonic :(
 // See https://github.com/hyperium/tonic/issues/1056
 #![allow(clippy::derive_partial_eq_without_eq)]
-
-//! A GRPC client to query containerd's API.
 
 pub use tonic;
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - run: cargo clippy --all-targets -- -D warnings
       - run: cargo clippy --all-targets --all-features -- -D warnings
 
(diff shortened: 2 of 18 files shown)


jsturtevant:
the failure is #207, should we disable that test to unblock the PR's?  Not sure why it didn't seem to be an issue on #181

crates/client/src/lib.rs
```diff
@@ -14,12 +14,11 @@
    limitations under the License.
 */
 
+#![cfg_attr(feature = "docs", doc = include_str!("../README.md"))]
```
jsturtevant, line 17: does this need the `//!`` so it formats properly?

jsturtevant (commented):
I looked but I didn't see where the docs get compiled on a PR, is it `cargo doc` that passes the feature?

crates/client/src/lib.rs
mxpv: No, we can include markdown directly here.
You can verify the final result with:
```bash
cargo doc --no-deps --features docs --open
```

mxpv:
> I looked but I didn't see where the docs get compiled on a PR, is it cargo doc that passes the feature?


I've updated CI to verify readmes on CI.

The `docs` feature is enabled when publishing to docs.rs via

```toml
[package.metadata.docs.rs]
 features = ["docs"]
```

crates/client/src/lib.rs
jsturtevant: cool!

PR #228, opened 2023-12-21
Return Container object in async `Runc::state()`

andrewbaxter:
It's a very small change so I figured it's simpler to open a PR than an issue first.

The sync `state` method returns `Container` but for async returns `Vec<usize>`, and I couldn't locate an explanation for why these might be different so I assume it's a mistake.  From a user perspective too I want Container rather than a usize vec.


mxpv (approved):

Burning1020 (approved):
LGTM

Issue #227, opened 2023-12-18
Is there a plan for  for registering and reporting cgroup OOM events in Rust shim?


aa624545345:
Go shim uses memory.oom_control and cgroup.event_control to register and report OOM (Out of Memory) events in the cgroup. Do you now have any plans for this aspect in Rust shim?

Burning1020:
Sounds good to me. I can handle this in this week.

Burning1020:
@aa624545345 Are you using shim with async feature?

aa624545345:
Not yet. I just tested the async version and found this lack. I commit a pr for this lack in #229

Burning1020:
> Not yet. I just tested the async version and found this lack. I commit a pr for this lack in #229

Oh, thank you.

PR #229, opened 2023-12-29
add oom monitor for rustshim

aa624545345:
Registering and reporting cgroup OOM events for Rust shim。 #227


crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,134 @@
...
+) -> Result<Receiver<String>> {
+    let path = cg_dir.join(event_name);
+    let event_file = fs::File::open(path.clone())
+        .await
+        .map_err(other_error!(e, "Error get path:"))?;
+
+    let eventfd = unsafe { libc::eventfd(0, libc::EFD_CLOEXEC) };
+
+    let event_control_path = cg_dir.join("cgroup.event_control");
+    let data = if arg.is_empty() {
```
Burning1020, line 100: `arg` is always empty, isn't it?

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,134 @@
...
+    };
+    fs::write(&event_control_path, data.clone())
+        .await
+        .map_err(other_error!(e, "Error write eventfd:"))?;
+
+    let mut buf = [0u8; 8];
+
+    let (sender, receiver) = mpsc::channel(128);
+    let key = key.to_string();
+    let mut eventfd_file = unsafe { File::from_raw_fd(eventfd) };
```
Burning1020, line 113: Same to nix Epoll

Burning1020 (changes requested):
Thanks for your pr, some of the methods already exist, so please change your code, thanks.

crates/runc-shim/src/task.rs
```diff
@@ -164,6 +188,23 @@ where
...
+                    .await
+                {
+                    Ok(rx) => rx,
+                    Err(e) => {
+                        return Err(ttrpc::Error::Others(format!(
+                            "notify_on_oom_v1 failed: {}",
+                            e
+                        )))
+                    }
+                };
```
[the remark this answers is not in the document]
aa624545345: Ok, i'll optimize the error handling here.

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,134 @@
...
+    cg_dir: &Path,
+    event_name: &str,
+    arg: &str,
+) -> Result<Receiver<String>> {
+    let path = cg_dir.join(event_name);
+    let event_file = fs::File::open(path.clone())
+        .await
+        .map_err(other_error!(e, "Error get path:"))?;
+
+    let eventfd = unsafe { libc::eventfd(0, libc::EFD_CLOEXEC) };
```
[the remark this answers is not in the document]
aa624545345: Initially, I was using the eventfd() function from the nix library. However, when transferring the ownership of the returned OwnFd value to tokio::spawn(), the event fd was being released. So I switched to using the rwa fd here.

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,134 @@
...
+    let s0_values: Vec<_> = s_values[0].trim().split(' ').collect();
+    if s0_values.len() < 6 {
+        return Err("Parse from mountinfo failed")
+            .map_err(other_error!(e, "Error get path from montinfo:"))?;
+    }
+    Ok((s0_values[3].to_string(), s0_values[4].to_string()))
+}
+
+pub async fn register_memory_event_v1(key: &str, dir: &Path) -> Result<Receiver<String>> {
+    register_memory_event(key, dir, "memory.oom_control", "").await
```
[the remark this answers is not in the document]
aa624545345: This PR only implements cgroup v1 for now, which is done to facilitate the future implementation of cgroup v2 and othher events. Or directly invoke the API, considering cgroup v2 and other events in the future.

crates/runc-shim/src/cgroup_memory.rs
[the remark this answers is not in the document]
aa624545345: I previously used the nix library, but encountered a problem with the eventfd being released. I can provide the reproduction steps later.

crates/runc-shim/src/container.rs
```diff
@@ -82,6 +84,8 @@ pub struct ContainerTemplate<T, E, P> {
     pub process_factory: P,
     /// exec processes of this container
     pub processes: HashMap<String, E>,
+    // memory cgroup path of this container
+    pub mem_cgruop_path: (String, String),
```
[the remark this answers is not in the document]
Burning1020: Yes, fetch and register it together.

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,134 @@
...
+use std::path::Path;
+use tokio::{
+    fs::{self, File},
+    io::{AsyncBufReadExt, AsyncReadExt, BufReader},
+    sync::mpsc::{self, Receiver},
+};
+
+use std::os::unix::io::{AsRawFd, FromRawFd};
+
+pub async fn get_path_from_cgorup(pid: u32) -> Result<String> {
```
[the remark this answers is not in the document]
Burning1020: I see, use yours if they don't have a async method.

crates/runc-shim/src/cgroup_memory.rs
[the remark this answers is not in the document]
Burning1020: > I previously used the nix library, but encountered a problem with the eventfd being released. I can provide the reproduction steps later.

Is the problem of `as_raw_fd()` and `into_raw_fd()`?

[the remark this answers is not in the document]
aa624545345: In the 0.27 version of nix, the return type has been changed to Ownfd. eventfd being released can be resolved using into_raw_fd. In my case, I directly move the ownership of the eventfd into the spawn.

aa624545345:
@mxpv @Burning1020  Hello, are there any further plans for this submission?

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,128 @@
...
+        if line.contains("cgroup") && line.contains("memory") {
+            return parse_memory_mountroot(&line);
+        }
+    }
+
+    Err("No memory found in mountinfo file.")
+        .map_err(other_error!(e, "Error get path from montinfo:"))?
+}
+
+fn parse_memory_mountroot(line: &str) -> Result<(String, String)> {
```
Burning1020, line 68: I understand you want to get the memory subsystem cgroup, how about using `CgroupBuilder::new().set_specified_controllers()`? It will give you a `MemoryResorecsBuilder` that implements getter and setter.

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,128 @@
...
+        .await
+        .map_err(other_error!(e, "Error write eventfd:"))?;
+
+    let mut buf = [0u8; 8];
+
+    let (sender, receiver) = mpsc::channel(128);
+    let key = key.to_string();
+
+    tokio::spawn(async move {
+        let mut eventfd_file = unsafe { File::from_raw_fd(eventfd.as_raw_fd()) };
```
Burning1020, line 108: Can you find another method to wrapper fd as a File? I think nix crate has a lot of safe function. Use unsafe function if there is only one choice.

crates/runc-shim/src/task.rs
```diff
@@ -93,13 +119,29 @@ impl<F, C> TaskService<F, C> {
             .await
             .unwrap_or_else(|e| warn!("send {} to publisher: {}", topic, e));
     }
+
+    fn run_oom_monitor(&self, mut rx: Receiver<String>, id: String) {
```
Burning1020, line 123: Sorry, I think your previous commit is better: https://github.com/containerd/rust-extensions/commit/9d1ccc9fc8e5998b562130c6cc3e03d91a4cf72b#diff-91a6ff7d9d978824549619245220f7b4913562d009ff1c7a78502141d063421bR113, `send_event()` need execute in another thread but cloning the whole TaskService is not smart.

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    Err("No memory found in mountinfo file.")
+        .map_err(other_error!(e, "Error get path from montinfo:"))?
+}
+
+fn parse_memory_mountroot(line: &str) -> Result<(String, String)> {
+    let s_values: Vec<_> = line.split(" - ").collect();
+    if s_values.len() != 2 {
```
mxpv, line 70: Let's replace this with `split_once`: https://doc.rust-lang.org/std/primitive.str.html#method.split_once

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,128 @@
...
+    let mut lines = BufReader::new(file).lines();
+    while let Some(line) =
+        lines
+            .next_line()
+            .await
+            .map_err(io_error!(e, "read file {} failed.", proc_path))?
+    {
+        if line.contains("memory") {
+            let parts: Vec<&str> = line.split(':').collect();
+            if let Some(path) = parts.last() {
```
mxpv, line 29: I believe there is no need to allocate `Vec` here?
Just `line.split(':').last()` should use iterator directly?

crates/runc-shim/src/cgroup_memory.rs
aa624545345: The reason why i don't use MemoryResorecsBuilder is because cgroup-rs lib doesn't save the third colmn of ‘/proc/self/mountinfo‘, and use it to trim paths get form `/proc/.../cgroup`:  [parse_mountinfo_for_line](https://github.com/kata-containers/cgroups-rs/blob/main/src/hierarchies.rs#L48).
This issue affects the correct retrieval of cgroup paths when running a container whithin a container.
Futuermore, impletention in cgroup-rs is not async when reading files  `/proc/.../cgroup` and `/proc/.../mountinfo`

crates/runc-shim/src/task.rs
aa624545345: Yes, I also agree with this. I will revert this part.

crates/runc-shim/src/cgroup_memory.rs
Burning1020: OK. Container in container is not a common scenarios.

crates/runc-shim/src/cgroup_memory.rs
aa624545345: I find two save functions in nix lib, [Epoll](https://docs.rs/nix/0.27.1/src/nix/sys/epoll.rs.html#105-189) and [read](https://docs.rs/nix/0.27.1/src/nix/unistd.rs.html#1107-1113). 
For Epoll, in this scenario, only one eventfd needs to be monitored, it seems that read directly seems more lightweight.
For read func in lib nix, its implementation is alse unsfe:
```rust
pub fn read(fd: RawFd, buf: &mut [u8]) -> Result<usize> {
    let res = unsafe {
        libc::read(fd, buf.as_mut_ptr() as *mut c_void, buf.len() as size_t)
    };

    Errno::result(res).map(|r| r as usize)
}
```
In addition, these two funcs are both sync.

crates/runc-shim/src/cgroup_memory.rs
aa624545345: Thx for your review, i'll modiry this.

crates/runc-shim/src/cgroup_memory.rs
aa624545345: Ok.

crates/runc-shim/src/cgroup_memory.rs
aa624545345: @Burning1020 Hi, i don't find a suitable way in lib nix to wrap a fd as a file. I just find two ways to avoid using `unsafe`, comparation is discribed above, using `unsafe { File::from_raw_fd())` seems better. Take a look here?

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use nix::sys::eventfd;
+use std::os::unix::io::{AsRawFd, FromRawFd};
+
+pub async fn get_path_from_cgorup(pid: u32) -> Result<String> {
+    let proc_path = format!("/proc/{}/cgroup", pid);
+    let file = File::open(proc_path.clone())
+        .await
+        .map_err(io_error!(e, "open {}.", proc_path))?;
+    let mut lines = BufReader::new(file).lines();
```
mxpv, line 20: Let's make this code a bit more Rusty:

```rust
    let proc_path = format!("/proc/{}/cgroup", pid);
    let (_, path) = read_to_string(&proc_path)
        .await
        .map_err(io_error!(e, "open {}.", proc_path))?
        .lines()
        .find(|line| line.contains("memory")) // Only one line contains "memory"
        .ok_or(Error::Other("Memory line not found".into()))?
        .split_once(":memory:") // Split by ":memory:"
        .ok_or(Error::Other("Failed to parse memory line".into()))?;

    Ok(path)
```

(I haven't tested it)

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    Err("No memory found in mountinfo file.")
+        .map_err(other_error!(e, "Error get path from montinfo:"))?
+}
+
+fn parse_memory_mountroot(line: &str) -> Result<(String, String)> {
+    let s_values = line
+        .split_once(" - ")
```
mxpv, line 69: I don't think we need to split the line by `-`, and then by whitespace.
We're interested in the first block anyway, so something like this should do?


```rust
    let mut lines = line.split_whitespace();

    let root = lines.nth(3).ok_or(Error::Other("...".into()))?;
    let mount = lines.next(). ok_or(Error::Other("...".into()))?;

    return (root.to_string(), mount.to_string())
```

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,126 @@
...
+    let mut buf = [0u8; 8];
+
+    let (sender, receiver) = mpsc::channel(128);
+    let key = key.to_string();
+
+    tokio::spawn(async move {
+        let mut eventfd_file = unsafe { File::from_raw_fd(eventfd.as_raw_fd()) };
+        loop {
+            match eventfd_file.read(&mut buf).await {
+                Ok(bytes_read) => {
```
mxpv, line 109:
```rust
Ok(bytes_read) if bytes_read == 0 => { ... }
```

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,126 @@
...
+                return Ok(path.to_string());
+            }
+            break;
+        }
+    }
+    Err("Memory cgroup path not found").map_err(other_error!(e, "Error get path:"))?
+}
+
+pub async fn get_existing_cgroup_mem_path(pid_path: String) -> Result<(String, String)> {
+    let mut mountinfo_path = get_path_from_mountinfo().await?;
```
mxpv, line 38: Let's use more meaningful names instead of `mountinfo_path.0` and `mountinfo_path.1`.
You can use the following syntax:
```rust
let (mut abc, cdb) = ...
```

mxpv (commented):
Left a few suggestions.
Also I'd love to see some unit tests if possible.

crates/runc-shim/src/cgroup_memory.rs
Burning1020: > @Burning1020 Hi, i don't find a suitable way in lib nix to wrap a fd as a file. I just find two ways to avoid using `unsafe`, comparation is discribed above, using `unsafe { File::from_raw_fd())` seems better. Take a look here?

No prob.

Burning1020:
> Left a few suggestions. Also I'd love to see some unit tests if possible.

Here I suggest the UT could be: 
1. mock a remote publisher
2. create a temp memory cgroup and set the limit, like 10K
3. keep watching this cgroup path
4. fork a child process that would apply for high memory, like 11K
```rust
match fork().unwrap() {
            ForkResult::Parent { child } => {
                let pid = child.as_raw() as u64;
                cg.add_task_by_tgid(CgroupPid::from(pid)).unwrap(); // add child to cgroup
                waitpid(Some(child), None).unwrap();
            }
            ForkResult::Child => {
                std::thread::sleep(Duration::from_millis(100));
                let mut v: Vec<u8> = vec![];
                v.resize(11 * 1024 * 1024, 0);
                exit(0);
            }
        };
```
5. make sure OOM event received
6. reasource cleanup

crates/runc-shim/src/cgroup_memory.rs
aa624545345: Alright, I’ll give it a try and make the necessary modifications.

crates/runc-shim/src/cgroup_memory.rs
aa624545345: The reson i used `-` to split the sting is because i referred to some libs for parsing the mountinfo file, using it to verify the validation of information about the mountinfo file. But i found it unnecessary using `-` after checking different platforms and compared with go shim, so here i will make some modifications here.

aa624545345:
> > Left a few suggestions. Also I'd love to see some unit tests if possible.
> 
> Here I suggest the UT could be:
> 
> 1. mock a remote publisher
> 2. create a temp memory cgroup and set the limit, like 10K
> 3. keep watching this cgroup path
> 4. fork a child process that would apply for high memory, like 11K
> 
> ```rust
> match fork().unwrap() {
>             ForkResult::Parent { child } => {
>                 let pid = child.as_raw() as u64;
>                 cg.add_task_by_tgid(CgroupPid::from(pid)).unwrap(); // add child to cgroup
>                 waitpid(Some(child), None).unwrap();
>             }
>             ForkResult::Child => {
>                 std::thread::sleep(Duration::from_millis(100));
>                 let mut v: Vec<u8> = vec![];
>                 v.resize(11 * 1024 * 1024, 0);
>                 exit(0);
>             }
>         };
> ```
> 
> 5. make sure OOM event received
> 6. reasource cleanup

Thanks for your comments.

aa624545345:
@Burning1020 @mxpv Hello, there are two points that i]m not quite sure about this PR:
1. This commit only works on cgroup v1, i think it is necessary to add detection for thel cgroup version of the rutime enviroment.(Impelemention of cgroup v2 canbe supported in a future update).
2. Regarding the mentioned UT tests, should i commit it in another PR?

Burning1020:
> @Burning1020 @mxpv Hello, there are two points that i]m not quite sure about this PR:
> 
> 1. This commit only works on cgroup v1, i think it is necessary to add detection for thel cgroup version of the rutime enviroment.(Impelemention of cgroup v2 canbe supported in a future update).
> 2. Regarding the mentioned UT tests, should i commit it in another PR?

1. Yes, I agree, cgroup v2 can be supported in future.
2. It's better to keep UT with your code in the same PR.

aa624545345:
@Burning1020 @mxpv @kzys hello, commit containers ut is pushed.

crates/runc-shim/src/task.rs
```diff
@@ -164,6 +191,23 @@ where
                 ..Default::default()
             })
             .await;
+            std::thread::sleep(std::time::Duration::from_secs(30));
```
Burning1020, line 194: Why sleep here?

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,184 @@
...
+
+use nix::sys::eventfd;
+use std::os::unix::io::{AsRawFd, FromRawFd};
+
+pub async fn get_path_from_cgorup(pid: u32) -> Result<String> {
+    let proc_path = format!("/proc/{}/cgroup", pid);
+    let path_string = read_to_string(proc_path.clone()).await.map_err(io_error!(
+        e,
+        "open {}.",
+        proc_path.clone()
```
mxpv, line 20: you can borrow `&proc_path` in both places instead of cloning.

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,184 @@
...
+            "Lines containers cgroup and memory not found in mountinfo".into(),
+        ))?;
+
+    parse_memory_mountroot(line)
+}
+
+fn parse_memory_mountroot(line: &str) -> Result<(String, String)> {
+    let mut columns = line.split_whitespace();
+    columns.nth(2);
+    let mount_root = columns.next().ok_or(Error::Other(
```
mxpv, line 61: why do we need `nth(2)` and `next` ? why not just `nth(3)` ?

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+    let mount_root = pid_path.trim_start_matches(&mount_root).to_string();
+    Ok((mount_root, mount_point))
+}
+
+async fn get_path_from_mountinfo() -> Result<(String, String)> {
+    let mountinfo_path = "/proc/self/mountinfo".to_string();
+    let mountinfo_string = read_to_string(mountinfo_path.clone())
+        .await
+        .map_err(io_error!(e, "open {}.", mountinfo_path.clone()))?;
```
mxpv, line 46:
```rust
     let mountinfo_path = "/proc/self/mountinfo";
     let mountinfo_string = read_to_string(mountinfo_path)
         .await
         .map_err(io_error!(e, "open {}.", mountinfo_path))?;
```

crates/runc-shim/src/task.rs
aa624545345: oh Sorry, this is for debugging, i'll remove this line.

crates/runc-shim/src/cgroup_memory.rs
aa624545345: Ok.

crates/runc-shim/src/cgroup_memory.rs
aa624545345: Or i can modify here as follows:
```rust
    let mount_root = columns.nth(3).ok_or(Error::Other(
        "Invalid input information about mountinfo".into(),
    ))?;
    let mount_point = columns.next().ok_or(Error::Other(
        "Invalid input information about mountinfo".into(),
    ))?;
```

aa624545345:
@mxpv @Burning1020 Revised according to the feedback, please review, thanks.

crates/runc-shim/src/cgroup_memory.rs
```diff
@@ -0,0 +1,182 @@
+use containerd_shim::{
```
Burning1020, line 1: Lisense header is needed! Refering: https://github.com/containerd/rust-extensions/blob/main/crates/runc-shim/src/common.rs#L1-L15

mxpv:
CI needs some work.
Other than that, looks good.

Burning1020 (approved):
LGTM. Thanks

aa624545345:
@Burning1020 @mxpv 
To fix CI, make the following modifications:
1. Use `fmt` to format code;
2. Add conditional compilation `#[cfg(target_os = "linux")]`;
3. Handling this error scenario, when registering oom monitor, trying to query the cgroup information of the container's main process, if the container has already exited, get_path_from_cgorup can't find the cgroup information of this container's main process, then this line will return the error directly:
```rust
let path_from_cgorup = cgroup_memory::get_path_from_cgorup(resp.pid).await?;
```
Modify the logic of registring the oom monitor to, during the construction of oom monitor, when error happens, not return, but instead log the error msgs.

mxpv (approved):

aa624545345:
@mxpv @Burning1020 Thx for the review, is cgroup2 version of oom monitor needed?

mxpv:
@aa624545345 what would be nice to have!

PR #230, opened 2024-01-02
"Subsystem::Mem"  add judge for the limit and swap

jokemanfire:
when update limit and swap will ocurr error.
this error wiil ocurr in  containerd's "TestUpdateContainerResources_MemoryLimit" test.


mxpv (approved):

Burning1020 (approved):
Sounds good. LGTM.

PR #232, opened 2024-01-05
The error will occur in containerdV1.7  syncexec function

jokemanfire:
you can get this error  easliy  when you run TestContainerDrainExecIOAfterExit function
ps: should kill this container when the process just leave a init process?


jokemanfire:
use false branch

PR #234, opened 2024-01-07
The golang "true" is 0 ,but Rust "true" is 1

jokemanfire:
The int variable value of the bool variable in Golang language is different from that of rust。
In golang 
// bool is the set of boolean values, true and false.
type bool bool

// true and false are the two untyped boolean values.
const (
	true  = 0 == 0 // Untyped bool.
	false = 0 != 0 // Untyped bool.
)

### DIFF
--- a/crates/runc-shim/src/common.rs
+++ b/crates/runc-shim/src/common.rs
@@ -248,3 +248,15 @@ pub(crate) fn xdg_runtime_dir() -> String {
     env::var("XDG_RUNTIME_DIR")
         .unwrap_or_else(|_| env::temp_dir().to_str().unwrap_or(".").to_string())
 }
+/// Golang Bool convert to Rust Bool
+pub struct Golang_Bool {
+    pub value: i32,
+}
+impl Golang_Bool {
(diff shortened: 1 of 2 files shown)


mxpv:
I'm not sure I understand this.
`KillRequest` is protobuf generated structure and boolean `all` field.
Where do you see it inverted?

jokemanfire:
In 'task.rs' kill function  ,    there is    "  info!("Kill request for {:?}", req);"     this way can print result .     but when you send the KillRequest in containerd , 
when ctr set all "true"   but shim get false ,  I find it in test function "TestContainerDrainExecIOAfterExit " ， Maybe I was wrong, But I feel confused , ctr send true ，but shim get false . so I think golang deal with bool  "0" represent true .

Burning1020:
Can you provide more defails, such as the log msg of `info!("Kill request for {:?}", req);`? It has a long way between `ctr` and `task.kill()`, I think it is not a bug of rust shim.

jokemanfire:
this is containerd's problem I will bring issus  to containerd

PR #235, opened 2024-01-07
if delete the process but the process's io still used by other process.tell ctr io still in use by other process.

jokemanfire:
find this error in  ctr v1.7   "TestContainerDrainExecIOAfterExit" ,if ctr send 'delete' info to shim ,the shim return ok ,the ctr will think that IO is being closed. but this IO may still used by other process , may shim can't close it .

### DIFF
--- a/crates/runc-shim/src/common.rs
+++ b/crates/runc-shim/src/common.rs
@@ -24,7 +24,7 @@ use std::{
         unix::io::RawFd,
     },
     path::Path,
-    sync::Arc,
+    sync::Arc, process::Command,
 };
 
 use containerd_shim::{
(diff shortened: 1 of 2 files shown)


Burning1020:
When failed to delete, shim will return the underlaying error. If IO was used by another process, I think it is the underlaying runtime that should report this error instead of shim.

jokemanfire:
Do you mean to implement it in runc.rs？

Burning1020:
> Do you mean to implement it in runc.rs？

Not yet,  runC command should do this. You can try go-shim and check its behaviour. If go-shim doesn't report, neither of rs-shim.

jokemanfire:
> > Do you mean to implement it in runc.rs？
> 
> Not yet, runC command should do this. You can try go-shim and check its behaviour. If go-shim doesn't report, neither of rs-shim.

I feel confused, Can you provide a solution？the go shim use pipe+fifo , so they can close pipe and fifo will not be used by Daemon , rust-shim use fifo directly,  you can't close the io , it will still used by Daemon.this the between goshim and rustshim , there's any nice solution to resolve this problem ?

Burning1020:
I guess you have a main container that has opened other containers' IO in your case, right? And you expect the other containers shouldn't get removed when their IO is still in use.

jokemanfire:
sorry,I don't think I described it clearly. I find this problem in "TestContainerDrainExecIOAfterExit" from containerd V1.7.6 , it use ExecSync function , but rustshim can not pass this test.

Burning1020:
> sorry,I don't think I described it clearly. I find this problem in "TestContainerDrainExecIOAfterExit" from containerd V1.7.6 , it use ExecSync function , but rustshim can not pass this test.

OK. Let me check it.

Wabct:
Hi @jokemanfire，I mentioned the same problem in [#9568](https://github.com/containerd/containerd/issues/9568).I thought of a solution, query the process that took the pipe when deleting, and kill then all. But I haven't tested this yet. Can your modification pass “TestContainerDrainExecIOAfterExit”?

jokemanfire:
> Hi @jokemanfire，I mentioned the same problem in [#9568](https://github.com/containerd/containerd/issues/9568).I thought of a solution, query the process that took the pipe when deleting, and kill then all. But I haven't tested this yet. Can your modification pass “TestContainerDrainExecIOAfterExit”?

yes, I found your issus and I have mentioned it in rust-shim discussion [,](https://github.com/containerd/rust-extensions/discussions/238) this code can pass that test, I think you need't kill them ,because containerd will take these io .Another solution  (like go shim use pipe + fifo) .

jokemanfire:
Hello , @Burning1020  @Wabct Is there an elegant solution to this problem found？ or this issus has been found?Thank you for your reply....  0.0

Wabct:
您好，我是姚鑫，您的邮件已收到，我查阅后会尽快回复您。谢谢！

PR #244, opened 2024-02-16
Automate vendoring proto files from containerd

mxpv:
This PR adds `update-vendor.sh` script to `scripts/` directory to automate updating vendoring changes. For now the script must be ran manually.

```
# A simple bash script to synchronize proto files from containerd to vendor/ directories of
# each crate.
#
# VERSION specified containerd release that script will download to extract protobuf files.
#
# For each crate, the script expects a text file named `rsync.txt` in the crate's directory.
# The file should contain a list of proto files that should be synchronized from containerd.
```

ref: https://github.com/containerd/rust-extensions/pull/202


scripts/update-vendor.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# For each crate, the script expects a text file named `rsync.txt` in the crate's directory.
+# The file should contain a list of proto files that should be synchronized from containerd.
+
+VERSION="v1.7.13"
+
+set -x
+
+# Download containerd source code.
+gh release download $VERSION \
```
Burning1020, line 16: `gh` is not a frequently used command, what about `wget` or `curl`?

mxpv: Fixed in https://github.com/containerd/rust-extensions/pull/244/commits/0dc97adc5e8dbdd65fe11d56cfbc8b14a9b2e562

Burning1020 (approved):
LGTM

PR #245, opened 2024-02-22
Remove read_file_to_str helper

mxpv:
Use `fs::read_to_string` from stdlib/tokio instead.

### DIFF
--- a/crates/shim/src/error.rs
+++ b/crates/shim/src/error.rs
@@ -84,6 +84,9 @@ pub enum Error {
     #[error("Failed to send exit event: {0}")]
     Send(#[from] std::sync::mpsc::SendError<ExitEvent>),
 
+    #[error("Failed to read file: {0}")]
+    Io(#[from] std::io::Error),
+
     #[error("Other: {0}")]
     Other(String),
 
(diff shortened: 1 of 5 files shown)


crates/shim/src/error.rs
```diff
@@ -84,6 +84,9 @@ pub enum Error {
     #[error("Failed to send exit event: {0}")]
     Send(#[from] std::sync::mpsc::SendError<ExitEvent>),
 
+    #[error("Failed to read file: {0}")]
+    Io(#[from] std::io::Error),
```
jsturtevant, line 88: Will this type of error always be `failed to read a file`?

jsturtevant (approved):
LGTM

Burning1020 (changes requested):

crates/shim/src/error.rs
mxpv: I don't really like the requirement to map errors and use macros everywhere, like
`map_err(io_error!`. I feel like [`anyhow.Context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#example) would be a better fit here.

Burning1020: However, anyhow error is not easy to downcast if you need check the exact type of error

mxpv: Where exactly we need to downcast?

Burning1020: > Where exactly we need to downcast?

When you want to ignore some specific error.

PR #246, opened 2024-02-22
Consolidate windows and linux CI yamls

mxpv:
Consolidate windows and linux yaml files, so `Report required job statuses` job can track windows jobs.


.github/workflows/ci.yml
```diff
@@ -34,26 +34,48 @@ jobs:
         env:
           RUSTDOCFLAGS: -Dwarnings
 
+  # TODO: Merge this with the checks job above
```
jsturtevant, line 37: just a note that we might not ever be able to merge this with the above since some of the packages such as the runc-shim will never work on Windows. maybe there is some fancy things we can do to work around not building all the packages?

jsturtevant (approved):
LGTM

.github/workflows/ci.yml
mxpv: I want to explore with more in follow ups. I have a few options that I want to investigate deeper.

PR #252, opened 2024-03-12
Release 0.7.0 of Shim and 0.5.0 of client

jsturtevant:
There have been some dependency version increases, updates to the protobuf files and small improvements since last release.  

In runwasi, we are blocked on an update since the containerd client wasn't on prost-types 12.x: https://github.com/containerd/runwasi/pull/488

PR #253, opened 2024-03-13
Fix command fds in sync version of the shim

mxpv:
Looks like we need to apply the patch https://github.com/containerd/rust-extensions/pull/251 for sync version as well to fix publishing https://github.com/containerd/rust-extensions/pull/252


jsturtevant:
do we need to update CI so it catches this?

jsturtevant (approved):
LGTM

PR #254, opened 2024-03-14
fix(shim): rewrite cgroups v2 parsing logic

Mossaka:
this commit rewrites the cgroups v2 parsing logic in get_cgroup function which is used to fetch stats of a container. The reason for the rewrite was that in some cases the original logic would panic due to index of bound for parsing paths like

`0::/kubepods-besteffort-pod162385e5_7f69_4c38_ba9c_db0a8f02b35e.slice:cri-containerd:278a0aac1fff30dfbc41b4a32ba9de4519928fe7480213dba87aa1498838ef34`

we ran into this issue in deleting a spin container in the spin shim.

the rewrite replaces index access to properly propogate the error to the caller of the function and added a few unit tests for the parsing logic.

see more discussion https://github.com/spinkube/containerd-shim-spin/issues/39, https://github.com/containerd/runwasi/issues/418

### DIFF
--- a/crates/shim/src/cgroup.rs
+++ b/crates/shim/src/cgroup.rs
@@ -209,7 +209,7 @@ pub fn get_cgroups_v2_path_by_pid(pid: u32) -> Result<PathBuf> {
 // https://github.com/opencontainers/runc/blob/1950892f69597aa844cbf000fbdf77610dda3a44/libcontainer/cgroups/fs2/defaultpath.go#L83
 fn parse_cgroups_v2_path(content: &str) -> Result<PathBuf> {
     // the entry for cgroup v2 is always in the format like `0::$PATH`
-    // where 0 is the hierarchy ID, the controller name is ommit in cgroup v2
+    // where 0 is the hierarchy ID, the controller name is omitted in cgroup v2
     // and $PATH is the cgroup path
     // see https://docs.kernel.org/admin-guide/cgroup-v2.html
     let Some(path) = content.strip_prefix("0::") else {
--- a/crates/shim/src/cgroup.rs
+++ b/crates/shim/src/cgroup.rs
@@ -16,7 +16,11 @@
 
 #![cfg(target_os = "linux")]
 
-use std::{fs, io::Read, path::Path};
+use std::{
+    fs,
+    io::Read,
+    path::{Path, PathBuf},
+};
 
 use cgroups_rs::{
     cgroup::get_cgroups_relative_paths_by_pid, hierarchies, Cgroup, CgroupPid, MaxValue, Subsystem,
@@ -179,7 +183,7 @@ fn get_cgroup(pid: u32) -> Result<Cgroup> {
     let hierarchies = hierarchies::auto();
     let cgroup = if hierarchies.v2() {
         let path = get_cgroups_v2_path_by_pid(pid)?;
-        Cgroup::load(hierarchies, path.as_str())
+        Cgroup::load(hierarchies, path)
     } else {
         // get container main process cgroup
         let path = get_cgroups_relative_paths_by_pid(pid)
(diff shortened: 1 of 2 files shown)


Mossaka:
The cgroups v2 parsing logic is inspired by https://github.com/opencontainers/runc/blob/1950892f69597aa844cbf000fbdf77610dda3a44/libcontainer/cgroups/fs2/defaultpath.go#L83

but there is a critical difference: 

The runc cgroups parsing logic will actually fail to parse 

`0::/kubepods-besteffort-pod162385e5_7f69_4c38_ba9c_db0a8f02b35e.slice:cri-containerd:278a0aac1fff30dfbc41b4a32ba9de4519928fe7480213dba87aa1498838ef34`

because it does not take the whole path splitted by the colons. The runc parsing logic would only get `kubepods-besteffort-pod162385e5_7f69_4c38_ba9c_db0a8f02b35e.slice` which is not the actual cgroups path.

I was not sure if this is a bug in runc, or a bug on how spin container's cgroups path was constructed in the first place.

jsturtevant:
Couple of questions:

- Is the path with colons like that valid? 
- What is generating the path with colons and should this be fixed in the code generating the path to be more idiomatic?

radu-matei:
I can validate that with this patch, what we are seeing in https://github.com/spinkube/containerd-shim-spin/issues/39 is now fixed.

crates/shim/src/cgroup.rs
```diff
@@ -199,6 +189,37 @@ fn get_cgroup(pid: u32) -> Result<Cgroup> {
...
+        return Err(Error::Other(format!("invalid cgroup path: {}", content)));
+    }
+
+    if parts[0] == "0" && parts[1].is_empty() {
+        // Check if parts[2] starts with '/', remove it if present.
+        let path = parts[2].strip_prefix('/').unwrap_or(parts[2]);
+        return Ok(format!("/sys/fs/cgroup/{}", path));
+    }
+
+    Err(Error::Other("cgroup path not found".into()))
```
jprendes, line 220: **suggestion:** I think this is slightly clearer:
```rust
let Some(path) = content.strip_prefix("0::") else {
    return Err(Error::Other(format!("invalid cgroup path: {}", content)));
}

let Ok(path) = PathBuf::from(format!("/sys/fs/cgroup/{path}")).canonicalize() {
    return Err(Error::Other("cgroup path not found".into()));
}

Ok(path)
```

crates/shim/src/cgroup.rs
```diff
@@ -199,6 +189,37 @@ fn get_cgroup(pid: u32) -> Result<Cgroup> {
     Ok(cgroup)
 }
 
+/// Get the cgroups v2 path given a PID
+pub fn get_cgroups_v2_path_by_pid(pid: u32) -> Result<String> {
+    // todo: should upstream to cgroups-rs
+    let path = format!("/proc/{}/cgroup", pid);
+    let content = fs::read_to_string(path).map_err(io_error!(e, "read cgroup"))?;
+    let content = content.trim_end_matches('\n');
```
jprendes, line 197: **question:** What should this do in the case of multiple lines in the file? Not even sure that's possible though :-)
**suggestion:**
```rust
let content = content.lines().next().unwrap_or("");
```

jprendes (approved):
LGTM.
Left minor comments.

crates/shim/src/cgroup.rs
Mossaka: this actually makes it a bit hard to tests, for the time being, I will leave it as is for now. Is that fine?

Mossaka:
> Is the path with colons like that valid?

I think runc with Go impl can actually handle path with colons. So it should be valid. 

> What is generating the path with colons and should this be fixed in the code generating the path to be more idiomatic?

Maybe youki's libcgroups? I am not entirely sure

Mossaka:
Could you please take a look, @mxpv 🙏?

crates/shim/src/cgroup.rs
jprendes: Good point.
How about this?
```rust
let Some(content) = content.lines().next() else {
    return Err(Error::Other("empty cgroup path".into()));
};

let Some(path) = content.strip_prefix("0::") else {
    return Err(Error::Other(format!("invalid cgroup path: {}", content)));
};

let path = path.trim_start_matches('/');

Ok(PathBuf::from(format!("/sys/fs/cgroup/{path}")))
```

Mossaka: Done! Thanks

crates/shim/src/cgroup.rs
```diff
@@ -199,6 +193,34 @@ fn get_cgroup(pid: u32) -> Result<Cgroup> {
...
+    let Ok(path) = parse_cgroups_v2_path(content)?.canonicalize() else {
+        return Err(Error::Other(format!("cgroup path not found")));
+    };
+    Ok(path)
+}
+
+// https://github.com/opencontainers/runc/blob/1950892f69597aa844cbf000fbdf77610dda3a44/libcontainer/cgroups/fs2/defaultpath.go#L83
+fn parse_cgroups_v2_path(content: &str) -> Result<PathBuf> {
+    // the entry for cgroup v2 is always in the format like `0::$PATH`
+    // where 0 is the hierarchy ID, the controller name is ommit in cgroup v2
```
jsturtevant, line 212:
```suggestion
    // where 0 is the hierarchy ID, the controller name is omitted in cgroup v2
```

crates/shim/src/cgroup.rs
```diff
@@ -199,6 +193,34 @@ fn get_cgroup(pid: u32) -> Result<Cgroup> {
...
 }
 
+/// Get the cgroups v2 path given a PID
+pub fn get_cgroups_v2_path_by_pid(pid: u32) -> Result<PathBuf> {
+    // todo: should upstream to cgroups-rs
+    let path = format!("/proc/{}/cgroup", pid);
+    let content = fs::read_to_string(path).map_err(io_error!(e, "read cgroup"))?;
+    let content = content.lines().next().unwrap_or("");
+
+    let Ok(path) = parse_cgroups_v2_path(content)?.canonicalize() else {
```
jsturtevant, line 203: is there any concern about resolving symlinks or relative paths via canonicalize here?

jsturtevant: In the go implementation the path is cleaned here https://github.com/opencontainers/runc/blob/1950892f69597aa844cbf000fbdf77610dda3a44/libcontainer/cgroups/fs2/defaultpath.go#L48

jprendes: Testing is harder if the actual path has to exist. Maybe we can just normalize it instead of canonicalizing it, although rust's stdlib doesnt have a normalize method, that would require a 3rd party crate.

jprendes: Or do something like this:
https://github.com/containers/youki/blob/0b89e1d52089c8c328496dab02dcc2bcf5e1b53c/crates/libcontainer/src/utils.rs#L92

mxpv (approved):

mxpv:
Frankly speaking (in longer term) I think we should not carry groups related code in the `shim` crate. shim crate supposed to be the foundation for building new shim runtimes for containers, without carrying any implementation details.

crates/shim/src/cgroup.rs
Mossaka: If I understand the Go's CleanPath correctly, it's like normalizing the path but without verifying it exists in the filesystem, unlick `fs::canonicalize()` right?

jsturtevant: I think the key is:
```
//  "This ensures that a
// path resulting from prepending another path will always resolve to lexically
// be a subdirectory of the prefixed path."
```

Burning1020 (approved):
lgtm

crates/shim/src/cgroup.rs
jsturtevant: > In the go implementation the path is cleaned here https://github.com/opencontainers/runc/blob/1950892f69597aa844cbf000fbdf77610dda3a44/libcontainer/cgroups/fs2/defaultpath.go#L48

@Burning1020 @mxpv does this path need be cleaned? It is there in the go code but I am not sure if I completely understand why it is needed. If not we should be ok to merge this.

Mossaka:
I think there are two pending issues which are not resolved by this PR where I made an issue for each

PR #262, opened 2024-04-23
Enable type names for Protocol Buffer message types to support `Any` decoding.

tobz:
## Context

When using something like [`EventsClient::subscribe`](https://docs.rs/containerd-client/0.5.0/containerd_client/services/v1/events_client/struct.EventsClient.html#method.subscribe), the [`Envelope`](https://docs.rs/containerd-client/0.5.0/containerd_client/services/v1/struct.Envelope.html) value returned provides the event data (if any) behind [`prost_types::Any`](https://docs.rs/prost-types/latest/prost_types/struct.Any.html).

This value can be fallibly decoded to the underlying type (such as [`containerd_client::events::ContainerCreate`](https://docs.rs/containerd-client/0.5.0/containerd_client/events/struct.ContainerCreate.html) for a container create event) by using [`Any::to_msg<M>`](https://docs.rs/prost-types/latest/prost_types/struct.Any.html#method.to_msg), but _only_ if `M` implements [`Name`](https://docs.rs/prost/0.12.4/prost/trait.Name.html).

Currently, based on how `tonic_build` is used to do the codegen, the `Name` trait is not implemented for any of the generated message types, making decoding much harder to do.

## Solution

This PR is a simple change that simply provides _additional_ configuration in `build.rs`, enabling the generation of `Name` implementations for message types. [`Builder::compile_with_config`](https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html) simply re-applies any of its own configuration on top of the configuration value given, so we're essentially just starting with a base of "enable type `Name` generation" and whatever else comes after (via `tonic_build::configure()...`) gets applied on top.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ oci-spec = "0.6"
 os_pipe = "1.1"
 prctl = "1.0.0"
 prost = "0.12"
+prost-build = "0.12"
 prost-types = "0.12"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -31,6 +31,7 @@ tower = { workspace = true, optional = true }
 
 [build-dependencies]
 tonic-build.workspace = true
+prost-build.workspace = true
 
 [features]
 connect = ["tokio", "tower"]
--- a/crates/client/build.rs
+++ b/crates/client/build.rs
@@ -65,9 +65,12 @@ const FIXUP_MODULES: &[&str] = &[
 ];
 
 fn main() {
+    let mut config = prost_build::Config::new();
+    config.enable_type_names();
+
     tonic_build::configure()
         .build_server(false)
-        .compile(PROTO_FILES, &["vendor/"])
+        .compile_with_config(config, PROTO_FILES, &["vendor/"])
         .expect("Failed to generate GRPC bindings");
 
     for module in FIXUP_MODULES {
--- a/crates/client/examples/container_events.rs
+++ b/crates/client/examples/container_events.rs
@@ -47,7 +47,7 @@ async fn main() {
                                 // required by the `Any` type specification. We add it manually here so that `prost` can
                                 // properly decode the payload.
                                 if !payload.type_url.starts_with('/') {
-                                    payload.type_url.insert_str(0, "/");
+                                    payload.type_url.insert(0, '/');
                                 }
 
                                 let payload: ContainerCreate = payload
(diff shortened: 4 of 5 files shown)


jsturtevant:
could you add an example in https://github.com/containerd/rust-extensions/tree/main/crates/client/examples that demonstrates this?  It would also ensure we have a way to validate it.

mxpv:
> It would also ensure we have a way to validate it.

I think a unit test with decoding should do it ?

Generally the PR is SGTM.

tobz:
I pushed both a unit test _and_ an example of listening for and decoding event payloads.

Admittedly, I have _not_ yet tested the example.

tobz:
Alright, I ended up testing this and it seems to work, but with one caveat.

You'll see in the example that I had to futz with the type URL field for the `Any` payloads. It appears that `containerd` generates non-compliant type URLs by not emitting a leading slash, so that a payload for the `ContainerCreate` event has a type URL of `containerd.events.ContainerCreate`, when `prost` generates the type URL as `/containerd.events.ContainerCreate`.

The Protocol Buffers definition for `Any` states that there must be _at least_ one forward slash in the type URL, which seems to be what `prost` is following and adhering to... but practically, it also seems pointless to have if there's no actual leading URL, given how all of this stuff was based on the premise of querying some schema/type registry endpoint and handling it all automagically... so I can understand why `containerd` just omitted the leading slash.

Either way, the functionality _does_ work, so long as we adjust the type URL to make `prost` happy. Hopefully that's sufficient.

Example output:

```text
toby@consigliera:~/src/rust-extensions$ cargo run --example container_events
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/examples/container_events`
container created: id=0f550a4e94ae12a8881cf017e14e3b505ea62b022f21876613490f99ce9549e6 payload=ContainerCreate { id: "0f550a4e94ae12a8881cf017e14e3b505ea62b022f21876613490f99ce9549e6", image: "", runtime: Some(Runtime { name: "io.containerd.runc.v2", options: Some(Any { type_url: "containerd.runc.v1.Options", value: [50, 4, 114, 117, 110, 99, 58, 28, 47, 118, 97, 114, 47, 114, 117, 110, 47, 100, 111, 99, 107, 101, 114, 47, 114, 117, 110, 116, 105, 109, 101, 45, 114, 117, 110, 99, 72, 1] }) }) }
container deleted: id=0f550a4e94ae12a8881cf017e14e3b505ea62b022f21876613490f99ce9549e6 payload=ContainerDelete { id: "0f550a4e94ae12a8881cf017e14e3b505ea62b022f21876613490f99ce9549e6" }
```

jsturtevant (approved):
Thanks!  I tried this out and it is working.

LGTM

mxpv (approved):

PR #264, opened 2024-04-24
fix(client): if tonic has tls feature

### LINKED ISSUE #261: containerd-client connect failed when toinc with tls feature
containerd-client connect failed when toinc with tls feature(Other deps need. )
```
called `Result::unwrap()` on an `Err` value: tonic::transport::Error(Transport, hyper::Error(Connect, HttpsUriWithoutTlsSupport(())))
```

example
```rust
use containerd_client::{connect};

async fn test()  {
    let channel = connect("/run/containerd/containerd.sock").await.unwrap();
}

#[tokio::main]
async fn main() {
    test().await;
}

```

Failed Cargo.toml
```
[dependencies]
tokio = { version = "1.37.0", features = ["full"] }
tower = "0.4"
tonic = { version = "0.11", default-features = false, features = ["default","transport", "codegen", "prost", "tls", "tls-roots", "gzip","tls-roots-common","tls-webpki-roots","channel"] }
containerd-client = { version="0.5" }

```

Success Cargo.toml
```
[dependencies]
tokio = { version = "1.37.0", features = ["full"] }
tower = "0.4"
tonic = { version = "0.11"}
containerd-client = { version="0.5" }

```

fcfangcc:
fixed #261 

In tonic, if uri startwith "https", will use tls connect.

https://github.com/hyperium/tonic/blob/eeb3268f71ae5d1107c937392389db63d8f721fb/tonic/src/transport/service/connector.rs#L74-L90

Just change uri prefix "http" can skip this logic.Like:
https://github.com/hyperium/tonic/blob/eeb3268f71ae5d1107c937392389db63d8f721fb/examples/src/uds/client.rs#L19

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -87,7 +87,9 @@ pub async fn connect(
 
     let path = path.as_ref().to_path_buf();
 
-    // Taken from https://github.com/hyperium/tonic/commit/b90c3408001f762a32409f7e2cf688ebae39d89e#diff-f27114adeedf7b42e8656c8a86205685a54bae7a7929b895ab62516bdf9ff252R15
+    // Taken from https://github.com/hyperium/tonic/blob/eeb3268f71ae5d1107c937392389db63d8f721fb/examples/src/uds/client.rs#L19
+    // There will ignore this uri because uds do not use it
+    // and make connection with UnixStream::connect.
     let channel = Endpoint::try_from("http://[::]")
         .unwrap()
         .connect_with_connector(tower::service_fn(move |_| {
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -88,7 +88,7 @@ pub async fn connect(
     let path = path.as_ref().to_path_buf();
 
     // Taken from https://github.com/hyperium/tonic/commit/b90c3408001f762a32409f7e2cf688ebae39d89e#diff-f27114adeedf7b42e8656c8a86205685a54bae7a7929b895ab62516bdf9ff252R15
-    let channel = Endpoint::try_from("https://[::]")
+    let channel = Endpoint::try_from("http://[::]")
         .unwrap()
         .connect_with_connector(tower::service_fn(move |_| {
             UnixStream::connect(path.clone())


crates/client/src/lib.rs
```diff
@@ -88,7 +88,7 @@ pub async fn connect(
     let path = path.as_ref().to_path_buf();
 
     // Taken from https://github.com/hyperium/tonic/commit/b90c3408001f762a32409f7e2cf688ebae39d89e#diff-f27114adeedf7b42e8656c8a86205685a54bae7a7929b895ab62516bdf9ff252R15
-    let channel = Endpoint::try_from("https://[::]")
+    let channel = Endpoint::try_from("http://[::]")
```
jsturtevant, line 91: are there some scenarios where we would want https?

mxpv: > tonic::transport::Error(Transport, hyper::Error(Connect, HttpsUriWithoutTlsSupport(())))

Possibly we could enable a missing feature in Tonic instead?

fcfangcc: The `https://[::]`  is disabled here, you can any str replace `https`.(Except https and tls feature enable)

Connect will  replace `connect_with_connector` function  with `xx.sock` .

Style of writing because Tonic channel only build with `Endpoint`

jsturtevant: Ok, I think I see now, the comment right above links to a sample that states:

```
// We will ignore this uri because uds do not use it
// if your connector does use the uri it will be provided
// as the request to the `MakeConnection`.
```

`Endpoint` parses as URI but the domain socket isn't a uri so we move on and connect to the with our client.  The uri that was initially passed is ignored in statement below `tower::service_fn(move |_<this is the uri>| {`

When the value is https and tls feature is enabled (as it is) it executes this code on  that temp value: https://github.com/hyperium/tonic/blob/eeb3268f71ae5d1107c937392389db63d8f721fb/tonic/src/transport/service/connector.rs#L80-L91

We could essentially change this string to anything we want that parses as an "endpoint" (Uri) like `unix://unused` since it isn't used.

We could also be more explicit and append `file:://` to the path like:

```
let channel = Endpoint::try_from("file://var/run/containerd/containerd.sock")
        .unwrap()
        .connect_with_connector(tower::service_fn(move |uri: tonic::transport::Uri| {
            let p = uri.path().to_owned();
            UnixStream::connect(p)
        }))
        .await?;
```

mxpv: oh, sorry, now I get it too :)
I think it deserves a comment.

jsturtevant:
LGTM, @fcfangcc could you add a comment explaining what is happening?

fcfangcc:
@jsturtevant You mean to add a comment explaining  in the source code?

jsturtevant:
[Yes](https://github.com/containerd/rust-extensions/pull/264#discussion_r1581729025), essentially describing a summary of https://github.com/containerd/rust-extensions/pull/264#discussion_r1581426817.  We had to dive in and dig around to figure out what was going on. It would be nice to leave a quick explanation for the next person to  point out what seems like a strange set up (setting up with http/https but actually ignoring that value all together and using our own logic to connect).

mxpv (approved):

PR #266, opened 2024-04-29
fix(overlay): If there are a large number of overlay layers, the container can't be started

zzzzzzzzzy9:
If there are a large number of overlay layers, the container can't be started. My image layer is about 60-80 layers.

### DIFF
--- a/crates/shim/src/mount.rs
+++ b/crates/shim/src/mount.rs
@@ -28,6 +28,8 @@ use log::error;
 #[cfg(target_os = "linux")]
 use nix::mount::{mount, MsFlags};
 #[cfg(target_os = "linux")]
+use nix::sched::{unshare, CloneFlags};
+#[cfg(target_os = "linux")]
 use nix::unistd::{fork, ForkResult};
 
 use crate::error::{Error, Result};
@@ -530,7 +532,15 @@ pub fn mount_rootfs(
     target: impl AsRef<Path>,
 ) -> Result<()> {
     //TODO add helper to mount fuse
-    //TODO compactLowerdirOption for overlay
+    let max_size = page_size::get();
+    // NOTE: 512 id a buffer during pagesize check.
+    let (chdir, options) =
+        if fs_type.unwrap_or("") == "overlay" && options_size(options) >= max_size - 512 {
+            LowerdirCompactor::new(options).compact()
+        } else {
+            (None, options.to_vec())
+        };
+
     let mut flags: MsFlags = MsFlags::from_bits(0).unwrap();
     let mut data = Vec::new();
     options.iter().for_each(|x| {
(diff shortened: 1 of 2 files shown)


zzzzzzzzzy9:
@mxpv

crates/shim/src/mount.rs
```diff
@@ -552,6 +560,11 @@ pub fn mount_rootfs(
         None
     };
 
+    if let Some(workdir) = chdir {
+        env::set_current_dir(Path::new(&workdir)).unwrap_or_else(|_| {
```
Burning1020, line 564: Changing the working directory of the current thread without changing back may pollute it. Should fork new process and change its working directory, otherwise, considering [Unshare(CLONE_FS)](https://github.com/containerd/containerd/pull/7513)

zzzzzzzzzy9: This function is called in tokio::task::spawn_blocking, spawn_blocking is a separate thread, not a worker in tokio, and tokio does not schedule other asynchronous functions to this thread when it is asynchronously scheduled, and which other asynchronous functions are only scheduled in the worker thread. The spawn_blocking thread is destroyed by Tokio every few seconds after the execution is complete. Is it possible to reduce the extra unnecessary overhead by not creating a new thread here? @Burning1020

Burning1020: Yes, I understand this function is running in another thread, but changing workding directory of one thread will affect another threads, because all threads in the same process having the same working directory. Here is an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=82ec578917c982ee95e37d5cd3a6ac26

zzzzzzzzzy9: Thank you for your suggestion, I will change this.

zzzzzzzzzy9:
@Burning1020 Done.

Burning1020 (approved):
LGTM. Have you ever test this in a real environment?

Burning1020 (approved):
LGTM. Have you ever test this in a real environment?

zzzzzzzzzy9:
Yes, I have tested in a 128-layers image and it works fine. @Burning1020

Burning1020:
> Yes, I have tested in a 128-layers image and it works fine. @Burning1020

OK. Thanks.

crates/shim/src/mount.rs
```diff
@@ -552,6 +562,12 @@ pub fn mount_rootfs(
         None
     };
 
+    unshare(CloneFlags::CLONE_FS).unwrap();
```
fuweid, line 565: why not just clone if we need to change current dir?

Burning1020: @zzzzzzzzzy9

ImSEten: That's right, it's my bad. I will change this.

PR #269, opened 2024-05-17
shim: add tracing macros

Mossaka:
This PR introduces `tracing` instrumentation to the shim crate. It will allow us to collect detailed function entry and exit, args and returns to understand the behaviour of the shim process. 

An use case for runwasi shim is that it will collect the traces and transform them to OTLP traces and output them to an jeager endpoint (see https://github.com/containerd/runwasi/pull/582). 

~~From my observation the perf overhead with tracing macros is minimal. If this is a concern, we can try disabling tracing staticlly. Statically, we may use rust conditional compilation feature to add either compile or ignore tracing macros that are added to each function.~~

Update: I've created a new `shim_instrument` proc-macro that will statically disable/enable tracing for each function. if the feature `tracing` is not enabled, the `instrument` macro is acts as a **no-opt**. Otherwise, it will apply `tracing::instrument` to that function.


Mossaka:
> Do we need to maintain one more crate for this? This seems to be achievable with:
> 
> ```rust
> #[cfg_attr(feature = "tracing", tracing::instrument)]
> ```

Ah yeah this looks like a better idea 👍.

crates/shim/src/args.rs
```diff
@@ -44,6 +44,7 @@ pub struct Flags {
 }
 
 /// Parses command line arguments passed to the shim.
+#[cfg_attr(feature = "tracing", tracing::instrument(parent = tracing::Span::current(), skip_all, level = "Info"))]
```
mxpv, line 47: If you do not specify the `parent` argument, the macro will by default use the currently active span as the parent. Seems like `parent = tracing::Span::current()` is not needed?

mxpv:
Same applies to level.

`Unless overridden, a span with the [INFO] will be generated.`

crates/shim/src/args.rs
Mossaka: Agreed. Will remove `parent = ...`. 

For `level = ...`, I actually prefer to explictly state `"Info"` since it's clearer what level of tracing are we collecting here.

mxpv (approved):

PR #270, opened 2024-05-24
monitor_notify_by_pid and monitor_unsubscribe may cause dead lock

zmrush:
monitor_notify_by_pid may be blocked by send msg to channel because bounded channel is full(but monitor lock has not be freed), on the other hand,monitor_unsubscribe may block because monitor lock is holded by the monitor_notify_by_pid

### DIFF
--- a/crates/shim/src/asynchronous/monitor.rs
+++ b/crates/shim/src/asynchronous/monitor.rs
@@ -19,7 +19,7 @@ use std::collections::HashMap;
 use lazy_static::lazy_static;
 use log::error;
 use tokio::sync::{
-    mpsc::{channel, Receiver, Sender},
+    mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
     Mutex,
 };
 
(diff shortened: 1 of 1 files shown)


zmrush:
@kzys @dims @caniszczyk @tianon hi,can anyone see this commit?thanks

Burning1020:
Can you find the reason why the monitor lock isn't being released? I think changing it to unbounded reduces the likelihood of this issue, but it doesn't resolve it.

zmrush:
monitor_notify_by_pid method need MONITOR lock firstly, and then send msg to channel which is bounded,therefore sending msg may be blocked,this causes the MONITOR lock not released.  Other methods may be like this: firstly get the lock then copy channels to a vector, then release the monitor lock, then send msg use the new built vector of channels, but it also maybe blocked because the bounded channel, of cause by this way the lock is released and there is no dead lock, but i think it also may be blocked because of the bounded channel. So i think change the bounded channel is always needed, which is implemented in the sync feature. @Burning1020

zmrush:
monitor_notify_by_pid method need MONITOR lock firstly, and then send msg to channel which is bounded,therefore sending msg may be blocked,this causes the MONITOR lock not released.  Other methods may be like this: firstly get the lock then copy channels to a vector, then release the monitor lock, then send msg use the new built vector of channels, but it also maybe blocked because the bounded channel, of cause by this way the lock is released and there is no dead lock, but i think it also may be blocked because of the bounded channel. So i think change the bounded channel is always needed, which is implemented in the sync feature. @Burning1020

Burning1020 (approved):
LGTM. Thanks

ningmingxiao:
I create another pr to fix this issue,we don't need to use  use unbounded_channel.

PR #271, opened 2024-05-24
FIFO may cause WouldBlock error

zmrush:
if FIFO stdin set non-block, process in container may open stdin by block method and it will cause WouldBlock error when it read from stdin

### DIFF
--- a/crates/runc/src/io.rs
+++ b/crates/runc/src/io.rs
@@ -336,6 +336,14 @@ impl Io for FIFO {
                 .read(true)
                 .custom_flags(libc::O_NONBLOCK)
                 .open(path)?;
+            unsafe {
+                let cur_flag = libc::fcntl(stdin.as_raw_fd(), libc::F_GETFL);
+                libc::fcntl(
+                    stdin.as_raw_fd(),
+                    libc::F_SETFL,
+                    cur_flag & !libc::O_NONBLOCK,
+                );
+            }
             cmd.stdin(stdin);
         }
 


zmrush:
@kzys @dims @caniszczyk @tianon hi,can anyone see this commit?thanks

zmrush:
@mxpv can u check this commit?

Burning1020:
Hi, @zmrush Do we have another solution? I'm not sure if opening it with O_NONBLOCK and then removing it afterward is a good approach.

zmrush:
E6A4E474-267D-4EBF-AB03-85BCCEBFF9FA

In fact,there is no fifo(Typed IO) in containerd, containerd has pipedio only and pipedio open stdin(pipe) with block flag. but our fifo open fd with non-block，which is inconsistent.  But if I remove the code(.custom_flags(libc::O_NONBLOCK)) of fifo and  open fifo-file with O_BLOCK,the set method of FIFO is blocked by our unit-test and unit-test failed. So i only can modify the O_NONBLOCK flag afterward like this commit. So if we can change our unit test, then we only remove the code (.custom_flags(libc::O_NONBLOCK)). If we do not change our unit test, I can only rewrite the code with this commit.  Wait for your reply

Burning1020:
> But if I remove the code(.custom_flags(libc::O_NONBLOCK)) of fifo and open fifo-file with O_BLOCK,the set method of FIFO is blocked by our unit-test and unit-test failed. 

That's also the reason why I add this flag at the first init. 
I just feel a little awkward, CC @mxpv @abel-von @jsturtevant

Issue #258, opened 2024-04-03
Remove cgroups from the shim crate


Mossaka:
Frankly speaking (in longer term) I think we should not carry cgroups related code in the `shim` crate. shim crate supposed to be the foundation for building new shim runtimes for containers on all platforms, without carrying any implementation details.

_Originally posted by @mxpv in https://github.com/containerd/rust-extensions/issues/254#issuecomment-2000413079_

PR #275, opened 2024-05-27
Feature: shim: Complete the metrics collection for cgroup v1

zzzzzzzzzy9:
Complete the metrics collection for cgroup v1 so that containerd can obtain complete metrics information

### DIFF
--- a/crates/shim/src/cgroup.rs
+++ b/crates/shim/src/cgroup.rs
@@ -23,19 +23,32 @@ use std::{
 };
 
 use cgroups_rs::{
-    cgroup::get_cgroups_relative_paths_by_pid, hierarchies, Cgroup, CgroupPid, MaxValue, Subsystem,
+    blkio::{BlkIoData, IoService},
+    cgroup::get_cgroups_relative_paths_by_pid,
+    hierarchies, Cgroup, CgroupPid, MaxValue, Subsystem,
 };
 use containerd_shim_protos::{
-    cgroups::metrics::{CPUStat, CPUUsage, MemoryEntry, MemoryStat, Metrics, PidsStat, Throttle},
+    cgroups::metrics::{
+        BlkIOEntry, BlkIOStat, CPUStat, CPUUsage, HugetlbStat, MemoryEntry, MemoryStat, Metrics,
+        PidsStat, Throttle,
+    },
     protobuf::{well_known_types::any::Any, Message},
     shim::oci::Options,
 };
+use log::debug;
 use oci_spec::runtime::LinuxResources;
 
 use crate::error::{Error, Result};
 
 // OOM_SCORE_ADJ_MAX is from https://github.com/torvalds/linux/blob/master/include/uapi/linux/oom.h#L10
 const OOM_SCORE_ADJ_MAX: i64 = 1000;
+const BLKIO_DEVICES_INFO_PATH: &str = "/proc/partitions";
+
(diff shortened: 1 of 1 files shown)


zzzzzzzzzy9:
@mxpv

zzzzzzzzzy9:
@Burning1020

crates/shim/src/cgroup.rs
```diff
@@ -102,15 +115,194 @@ fn write_process_oom_score(pid: u32, score: i64) -> Result<()> {
...
+                for line in stat.lines() {
+                    let parts = line.split(' ').collect::<Vec<&str>>();
+                    if parts.len() != 2 {
+                        Err(Error::Other(format!("invalid cpu stat line: {}", line)))?;
+                    }
+
+                    // https://github.com/opencontainers/runc/blob/dbe8434359ca35af1c1e10df42b1f4391c1e1010/libcontainer/cgroups/fs2/cpu.go#L70
+                    match parts[0] {
+                        "nr_periods" => {
+                            throttle.set_periods(parts[1].parse::<u64>().unwrap());
```
Burning1020, line 150: Could you please use `unwrap_or_default()` to avoiding panic directly?

crates/shim/src/cgroup.rs
```diff
@@ -184,6 +369,134 @@ pub fn collect_metrics(pid: u32) -> Result<Metrics> {
...
+                        if *s[0] == device_key.major.to_string()
+                            && *s[1] == device_key.minor.to_string()
+                        {
+                            return Ok(s[3].to_string());
+                        }
+                    }
+                    // if we don't find the device in /proc/partitions, just return a nil string.
+                    Ok("".to_string())
+                }
+                Err(e) => Err(Error::Other(
```
Burning1020, line 401: we have `other!()`

Mossaka:
I think it's great to have a complete cgroupv1 stats collection, but given https://github.com/containerd/rust-extensions/issues/258 suggests that the complex logic of cgroups stats should not reside in the shim crate, I would recommend that we should take a look at if [libcgroups](https://github.com/containers/youki/tree/main/crates/libcgroups) could be of use.

In particular, `libcgroups` maintained by youki has cgroups v1 and v2 `StatsProvider` ([source](https://github.com/containers/youki/blob/main/crates/libcgroups/src/v1/cpu.rs#L59-L87)), so we don't need to duplicate the effort in this repo. 

I am fully aware that we are using [cgroups-rs](https://github.com/kata-containers/cgroups-rs), so what I am proposing here is to evaluate whether or not `libcgroups` satisfies our requirements and can replace `cgroups-rs`.

CC @utam0k

utam0k:
I'm happy to collaborate with you if possible!

crates/shim/src/cgroup.rs
zzzzzzzzzy9: Done.

crates/shim/src/cgroup.rs
zzzzzzzzzy9: Done.

mxpv:
Hey @Mossaka, can you give a bit more context on `cgroups-rs` vs `libcgroups` ?

mxpv:
Thanks for your work on this! Since the long-term plan is to remove cgroups from the shim crate (see [#258](https://github.com/containerd/rust-extensions/issues/258)), I’m going to close this PR.

PR #277, opened 2024-05-30
move unshare into chdir

zzzzzzzzzy9:
We don't need clone if we don't need to change current dir.

### DIFF
--- a/crates/shim/src/mount.rs
+++ b/crates/shim/src/mount.rs
@@ -562,8 +562,8 @@ pub fn mount_rootfs(
         None
     };
 
-    unshare(CloneFlags::CLONE_FS).unwrap();
     if let Some(workdir) = chdir {
+        unshare(CloneFlags::CLONE_FS).unwrap();
         env::set_current_dir(Path::new(&workdir)).unwrap_or_else(|_| {
             unsafe { libc::_exit(i32::from(MountExitCode::ChdirErr)) };
         });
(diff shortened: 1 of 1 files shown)


zzzzzzzzzy9:
@Burning1020 @fuweid Thanks for suggestion.
https://github.com/containerd/rust-extensions/pull/266

crates/shim/src/mount.rs
```diff
@@ -562,8 +562,8 @@ pub fn mount_rootfs(
         None
     };
 
-    unshare(CloneFlags::CLONE_FS).unwrap();
     if let Some(workdir) = chdir {
+        unshare(CloneFlags::CLONE_FS).unwrap();
```
fuweid, line 566: Would you please add UT for that? Thanks

zzzzzzzzzy9: Yes, I will do it later.

zzzzzzzzzy9: Done.

crates/shim/src/mount.rs
```diff
@@ -723,4 +723,47 @@ mod tests {
             assert_eq!(options, expected_options);
         }
     }
+
+    #[test]
+    fn test_mount_rootfs() {
+        let fs_type = Some("overlay");
+        let source = Some("overlay");
+        let workdir =
+            "/var/lib/containerd-test/io.containerd.snapshotter.v1.overlayfs/snapshots/0/work";
```
fuweid, line 732: To be honest, I don't think it's good idea to create test folder like this. Please use `/tmp/$RANDOM-LONG-STRING`. Thanks.

crates/shim/src/mount.rs
```diff
@@ -723,4 +723,47 @@ mod tests {
...
+            "index=off".to_string(),
+            "workdir=".to_string() + workdir,
+            "upperdir=".to_string() + upperdir,
+            "lowerdir=".to_string() + dirs.join(":").as_str(),
+        ];
+
+        let target = "/run/containerd/io.containerd.runtime.v2.task/k8s.io/mount-test/rootfs";
+        std::fs::create_dir_all(target).unwrap();
+        let current_dir = env::current_dir().unwrap();
+        std::thread::spawn(move || mount_rootfs(fs_type, source, &options, target).unwrap());
```
fuweid, line 757: Why must it spawn thread to mount it?

zzzzzzzzzy9: because this mount_rootfs is called in a new thread by tokio::spawn_blocking, so I think it's better to test mount_rootfs in a spawn thread.

PR #285, opened 2024-06-24
shim: Implement the pause and resume interfaces

zzzzzzzzzy9:
Implement the `pause` and `resume` interfaces

### DIFF
--- a/crates/runc-shim/src/processes.rs
+++ b/crates/runc-shim/src/processes.rs
@@ -55,6 +55,8 @@ pub trait Process {
     async fn stats(&self) -> Result<Metrics>;
     async fn ps(&self) -> Result<Vec<ProcessInfo>>;
     async fn close_io(&mut self) -> Result<()>;
+    async fn pause(&mut self) -> Result<()>;
+    async fn resume(&mut self) -> Result<()>;
 }
 
 #[async_trait]
@@ -65,6 +67,8 @@ pub trait ProcessLifecycle<P: Process> {
     async fn update(&self, p: &mut P, resources: &LinuxResources) -> Result<()>;
     async fn stats(&self, p: &P) -> Result<Metrics>;
     async fn ps(&self, p: &P) -> Result<Vec<ProcessInfo>>;
+    async fn pause(&self, p: &mut P) -> Result<()>;
+    async fn resume(&self, p: &mut P) -> Result<()>;
 }
 
 pub struct ProcessTemplate<S> {
@@ -197,5 +201,13 @@ where
             drop(stdin_w_file);
         }
         Ok(())
+    }
+
+    async fn pause(&mut self) -> Result<()> {
+        self.lifecycle.clone().pause(self).await
+    }
+
+    async fn resume(&mut self) -> Result<()> {
+        self.lifecycle.clone().resume(self).await
     }
 }
--- a/crates/runc-shim/src/runc.rs
+++ b/crates/runc-shim/src/runc.rs
@@ -349,6 +349,46 @@ impl ProcessLifecycle<InitProcess> for RuncInitLifecycle {
             })
             .collect())
     }
+
+    #[cfg(target_os = "linux")]
+    async fn pause(&self, p: &mut InitProcess) -> Result<()> {
+        match p.state {
+            Status::RUNNING => {
+                p.state = Status::PAUSING;
+                if let Err(e) = self.runtime.pause(p.id.as_str()).await {
+                    p.state = Status::RUNNING;
+                    return Err(runtime_error(&self.bundle, e, "OCI runtime pause failed").await);
+                }
+                p.state = Status::PAUSED;
+                Ok(())
+            }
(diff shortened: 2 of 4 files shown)


zzzzzzzzzy9:
@Burning1020

crates/runc-shim/src/processes.rs
```diff
@@ -83,7 +87,7 @@ pub struct ProcessTemplate<S> {
 impl<S> ProcessTemplate<S> {
     pub fn new(id: &str, stdio: Stdio, lifecycle: S) -> Self {
         Self {
-            state: Status::CREATED,
+            state: tokio::sync::Mutex::new(Status::CREATED),
```
mxpv, line 90: Any reason why we require mutex? Seems like all functions are mutable.

crates/runc-shim/src/runc.rs
```diff
@@ -349,6 +349,56 @@ impl ProcessLifecycle<InitProcess> for RuncInitLifecycle {
...
+                if let Err(e) = self.runtime.pause(p.id.as_str()).await {
+                    *state = Status::RUNNING;
+                    return Err(runtime_error(&self.bundle, e, "OCI runtime pause failed").await);
+                }
+                *state = Status::PAUSED;
+                Ok(())
+            }
+            Status::STOPPED => Err(other!("cannot pause a stopped process")),
+            Status::PAUSED => Err(other!("cannot pause a paused process")),
+            Status::PAUSING => Err(other!("cannot pause a pausing process")),
```
mxpv, line 370: All non-running cases can be covered with:

```rust
_ => Err(other!("cannot pause when in {:?} state", *state))
```

crates/runc-shim/src/runc.rs
```diff
@@ -349,6 +349,56 @@ impl ProcessLifecycle<InitProcess> for RuncInitLifecycle {
...
+    async fn resume(&self, p: &mut InitProcess) -> Result<()> {
+        let mut state = p.state.lock().await;
+        match *state {
+            Status::UNKNOWN => Err(other!("cannot resume an unknown process")),
+            Status::CREATED => Err(other!("cannot resume a created process")),
+            Status::RUNNING => Err(other!("cannot resume a running process")),
+            Status::STOPPED => Err(other!("cannot resume a stopped process")),
+            Status::PAUSED => {
+                if let Err(e) = self.runtime.resume(p.id.as_str()).await {
+                    return Err(runtime_error(&self.bundle, e, "OCI runtime pause failed").await);
```
mxpv, line 389: `map_err` is more canonical.

crates/runc-shim/src/runc.rs
zzzzzzzzzy9: Yes, that's right, thanks

crates/runc-shim/src/runc.rs
zzzzzzzzzy9: `runtime_error` is an async function, cannot used by map_err.

crates/runc-shim/src/processes.rs
zzzzzzzzzy9: I think, the state should be atomic. State changes should be made by only one thread.

crates/runc-shim/src/runc.rs
mxpv: Right, but `.await` returns `Result`, so you can do `.await.map_err(...)`?

crates/runc-shim/src/processes.rs
mxpv: Correct, so Rust will enforce this at compile time.
Functions which have `&mut self` mutate state, so you won't be able to call them from multiple threads.

crates/runc-shim/src/runc.rs
zzzzzzzzzy9: Do you mean `self.runtime.resume(p.id.as_str()).await.map_err(runtime_error(...))?`?
This will not pass compilation because the `runtime_error` function is an async function and cannot be used in `map_err`

crates/runc-shim/src/processes.rs
zzzzzzzzzy9: Got it, thanks for your explanation.

zzzzzzzzzy9: Done.

mxpv:
CI needs some attention.

zzzzzzzzzy9:
CI is caused by protobuf-3.5.0. The maintainer of this library upgraded to version 3.5.0 a few days ago, resulting in the fact that it cannot be compiled even without my commit. Is there any solution? @mxpv

mxpv:
This was addressed in https://github.com/containerd/rust-extensions/pull/296
Could you pls rebase the PR?

mxpv:
ping @zzzzzzzzzy9 are you still interested to contribute this?

zzzzzzzzzy9:
yes ,i will rebase this pr

zzzzzzzzzy9:
Done. @mxpv

PR #290, opened 2024-07-11
add pid to the delete response

zzzzzzzzzy9:
add pid to the delete response

### DIFF
--- a/crates/shim/src/asynchronous/util.rs
+++ b/crates/shim/src/asynchronous/util.rs
@@ -87,6 +87,12 @@ pub async fn write_str_to_file(filename: impl AsRef<Path>, s: impl AsRef<str>) -
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


crates/shim/src/asynchronous/util.rs
```diff
@@ -87,6 +87,12 @@ pub async fn write_str_to_file(filename: impl AsRef<Path>, s: impl AsRef<str>) -
     Ok(())
 }
 
+pub async fn read_pid_from_file(pid_path: &Path) -> Result<i32> {
```
Mossaka, line 90: Question: Should we maintain parity between sync and async APIs? Specifically, should any additions or modifications to one module be reflected in the other?

zzzzzzzzzy9: That's right, but `read_pid_from_file` this function has been implemented in `synchronous`.

mxpv (approved):

PR #291, opened 2024-07-16
Unmount the mount point of the container when the container is deleted.

zzzzzzzzzy9:
Unmount the mount point of the container when the container is deleted.

Shim will mount the rootfs on host, so when container is deleted and the shim is dead, shim should unmount the mountpoint who has been mounted by shim.


mxpv:
I guess same suggestion here as in https://github.com/containerd/rust-extensions/pull/305#discussion_r1702070554. There is a lot of linux specific stuff. Let's move those to a separate file to have less linux targets?

PR #298, opened 2024-07-30
set worker thread nums through env variables

mingfukuang:
The current shim process can run on all cores in the system, which is not suitable for scenarios with high CPU load. Add a shim thread count setting for runtime, the number of shim worker threads can be controlled through environment variables.

### DIFF
--- a/crates/runc-shim/src/main.rs
+++ b/crates/runc-shim/src/main.rs
@@ -15,7 +15,7 @@
 */
 
 use std::env;
-
+use tokio::runtime::Builder;
 use containerd_shim::{asynchronous::run, parse};
 
 mod cgroup_memory;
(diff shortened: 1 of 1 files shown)


mxpv:
@mingfukuang If I understand this right, Tokio already supports this out of the box via `TOKIO_WORKER_THREADS` env variable.

`
The core threads are where all asynchronous code runs, and Tokio will by default spawn one for each CPU core. You can use the environment variable TOKIO_WORKER_THREADS to override the default value.
`

See https://docs.rs/tokio/latest/tokio/#cpu-bound-tasks-and-blocking-code and https://github.com/tokio-rs/tokio/pull/4250

mxpv:
Going to close this one, let me know if I missed anything.

mingfukuang:
@mxpv Agreed thank you for your explanation.

PR #305, opened 2024-08-01
support loop-dev mount in mount_rootfs

zzzzzzzzzy9:
support loop-dev mount in mount_rootfs

### DIFF
--- a/crates/shim/src/mount.rs
+++ b/crates/shim/src/mount.rs
@@ -22,6 +22,8 @@ use std::{
     ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, Not},
     path::Path,
 };
+#[cfg(target_os = "linux")]
+use std::{fs::File, os::fd::AsRawFd};
 
 use lazy_static::lazy_static;
 use log::error;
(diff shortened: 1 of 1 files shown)


crates/shim/src/mount.rs
```diff
@@ -615,6 +690,164 @@ pub fn mount_rootfs(
     Err(Error::Unimplemented("start".to_string()))
 }
 
+#[cfg(target_os = "linux")]
```
mxpv, line 693: I wonder if we should move linux specific code to `mount_linux.rs` ?
So we don't have to specify target for each function.

mxpv (commented):
It'd great to have some unit tests to cover loop dev functionality.

crates/shim/src/mount.rs
```diff
@@ -615,6 +690,164 @@ pub fn mount_rootfs(
...
+    }
+    unsafe {
+        #[cfg(target_env = "gnu")]
+        let ret = libc::ioctl(
+            loop_dev.as_raw_fd() as libc::c_int,
+            LOOP_SET_STATUS64 as libc::c_ulong,
+            info,
+        );
+        #[cfg(target_env = "musl")]
+        if let Err(e) = nix::errno::Errno::result(ret) {
```
Tim-Zhang, line 806: The variable ret does not defined in musl env

PR #313, opened 2024-09-12
Process not found should not return ok

jokemanfire:
if the process ID is not found, I think an error should be returned.

### DIFF
--- a/crates/runc-shim/src/container.rs
+++ b/crates/runc-shim/src/container.rs
@@ -136,7 +136,6 @@ where
         let process = self.get_mut_process(exec_id_opt);
         match process {
             Ok(p) => p.delete().await?,
-            Err(Error::NotFoundError(_)) => return Ok((pid, code, exited_at)),
             Err(e) => return Err(e),
         }
         if let Some(exec_id) = exec_id_opt {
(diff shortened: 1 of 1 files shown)


mxpv:
ping @Burning1020

jsturtevant:
> if the process ID is not found, I think an error should be returned.

Could you expand on why?  This is during a delete of the container.  If there is no process then it should be a noop.

jokemanfire:
Hello @jsturtevant   I want to align it with Goshim .  Return ok will ignore the Error::NotFoundError.
go shim code like this 
https://github.com/containerd/containerd/blob/main/cmd/containerd-shim-runc-v2/runc/container.go

`// Process returns the process by id
func (c *Container) Process(id string) (process.Process, error) {
	c.mu.Lock()
	defer c.mu.Unlock()
	if id == "" {
		if c.process == nil {
			return nil, fmt.Errorf("container must be created: %w", errdefs.ErrFailedPrecondition)
		}
		return c.process, nil
	}
	p, ok := c.processes[id]
	if !ok {
		return nil, fmt.Errorf("process does not exist %s: %w", id, errdefs.ErrNotFound)
	}
	return p, nil
}`

mxpv (approved):

PR #316, opened 2024-09-27
fix exec hang when tx channel is full

ningmingxiao:
@Burning1020  @mxpv another way to fix https://github.com/containerd/rust-extensions/pull/270, this pr don't need to use unbounded_channel
can you review this pr ? Thank you.

### DIFF
--- a/crates/runc-shim/src/service.rs
+++ b/crates/runc-shim/src/service.rs
@@ -159,52 +159,51 @@ async fn process_exits(
             if let Subject::Pid(pid) = e.subject {
                 debug!("receive exit event: {}", &e);
                 let exit_code = e.exit_code;
-                for (_k, cont) in containers.lock().await.iter_mut() {
-                    let bundle = cont.bundle.to_string();
-                    // pid belongs to container init process
-                    if cont.init.pid == pid {
-                        // kill all children process if the container has a private PID namespace
-                        if should_kill_all_on_exit(&bundle).await {
-                            cont.kill(None, 9, true).await.unwrap_or_else(|e| {
-                                error!("failed to kill init's children: {}", e)
-                            });
-                        }
-                        // set exit for init process
-                        cont.init.set_exited(exit_code).await;
+                let containers = containers.clone();
+                let tx = tx.clone();
+                tokio::spawn(async move {
+                    for (_k, cont) in containers.lock().await.iter_mut() {
+                        let bundle = cont.bundle.to_string();
+                        // pid belongs to container init process
+                        if cont.init.pid == pid {
+                            // kill all children process if the container has a private PID namespace
+                            if should_kill_all_on_exit(&bundle).await {
+                                cont.kill(None, 9, true).await.unwrap_or_else(|e| {
+                                    error!("failed to kill init's children: {}", e)
+                                });
+                            }
+                            // set exit for init process
+                            cont.init.set_exited(exit_code).await;
 
-                        // publish event
-                        let (_, code, exited_at) = match cont.get_exit_info(None).await {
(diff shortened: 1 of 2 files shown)


ningmingxiao:
@Burning1020  another way to fix https://github.com/containerd/rust-extensions/pull/270 don't need to use unbounded_channel

mxpv:
Could you pls rebase against latest main to fix the CI ?

ningmingxiao:
done

mxpv:
Trying to understand this change. Can you elaborate how exactly this fixes the issue?

You spawn an extra thread on each iteration and each new thread still writes to the channel. There are no consumers on this side, right? So why does this help?

crates/runc-shim/src/service.rs
```diff
@@ -159,52 +159,56 @@ async fn process_exits(
...
-                    for (_exec_id, p) in cont.processes.iter_mut() {
-                        // set exit for exec process
-                        if p.pid == pid {
-                            p.set_exited(exit_code).await;
-                            // TODO: publish event
                             break;
                         }
+
+                        // pid belongs to container common process
+                        for (_exec_id, p) in cont.processes.iter_mut() {
```
mxpv, line 202: nit:

```rust

    if let Some(p) = cont.processes.values_mut().find(|p| p.pid == pid) {
        p.set_exited(exit_code).await;
    }

```

ningmingxiao:
> Trying to understand this change. Can you elaborate how exactly this fixes the issue?
> 
> You spawn an extra thread on each iteration and each new thread still writes to the channel. There are no consumers on this side, right? So why does this help?

I add some log find that containers.lock() can't get lock when chan is full.(and I add some log I can't get who hold the lock),
So I split it into two tokio::spawn,consumer can consumer the channel quickly.
```
❯ cat container.json 
{
  "metadata": {
    "name": "gpu"
  },
  "image": {
    "image": "docker.io/library/busybox:1.28"
  },
  "command": [
    "top"
  ],
  "tty": true,
  "linux": {
  }
}

root in /home/nmx/pods via 🐹 v1.21.12 
❯ cat pod.json 
{
  "metadata": {
    "name": "sandbox-a10798895196912d15189a32112",
    "namespace": "default",
    "attempt": 1,
    "uid": "hdishd83djaidwnduwk28bcsb"
  },
  "log_directory": "/tmp",
  "linux": {
  }
}

crictl  run    --no-pull container.json  pod.json
```
main.go
```
package main

import (
        "fmt"
        "math/rand"
        "os/exec"
        "sync"
        "time"
)

func main() {
        test()
        fmt.Println("done")
}

func runCmd(command string) {
        cmd := exec.Command("bash", "-c", command)
        err := cmd.Run()
        if err != nil {
                fmt.Println(err)
        }
}

func test() {
        var wg sync.WaitGroup
        const numContainers = 500
        for i := 1; i <= numContainers; i++ {
                wg.Add(1)
                go func(i int) {
                        rand.Seed(time.Now().UnixNano())
                        randomNumber := rand.Intn(10) + 1
                        cmdcri := "crictl exec 2380a  sleep " + fmt.Sprintf("%d", randomNumber)
                        fmt.Println(cmdcri)
                        defer wg.Done()
                        runCmd(cmdcri)
                }(i)
        }
        wg.Wait()
}

```
and limit work_threads to 2 make it easier to reproduce
```
#[tokio::main(flavor = "multi_thread", worker_threads = 2)]
async fn main() {
    parse_version();
    run::<Service>("io.containerd.runc.v2-rs", None).await;
}
```

ningmingxiao:
ping @mxpv  @Burning1020

PR #317, opened 2024-09-29
support to build offline to fix too new version  break build

ningmingxiao:


mxpv:
I've pinned ttrpc versions. Going to close this one.

Issue #312, opened 2024-09-02
event publisher looks like inadequate


jokemanfire:
I want to add a queue in the publisher to try reporting messages and give a 5 seconds timeout, if necessary.

PR #318, opened 2024-09-30
<rshim>Improve the events report

jokemanfire:
1.add timeout in ttrpc ctx
2.add queue in publisher report
3.add reconnect client method

### DIFF
--- a/crates/runc-shim/src/service.rs
+++ b/crates/runc-shim/src/service.rs
@@ -27,12 +27,12 @@ use containerd_shim::{
     event::Event,
     io_error,
     monitor::{Subject, Topic},
-    protos::{events::task::TaskExit, protobuf::MessageDyn},
+    protos::{events::task::TaskExit, protobuf::MessageDyn, ttrpc::context::with_timeout},
     util::{
         convert_to_timestamp, read_options, read_pid_from_file, read_runtime, read_spec, timestamp,
         write_str_to_file,
     },
-    Config, Context, DeleteResponse, Error, Flags, StartOpts,
+    Config, DeleteResponse, Error, Flags, StartOpts,
 };
 use log::{debug, error, warn};
 use tokio::sync::mpsc::{channel, Receiver, Sender};
@@ -218,8 +218,11 @@ async fn forward(
 ) {
     tokio::spawn(async move {
         while let Some((topic, e)) = rx.recv().await {
+            // While ttrpc push the event,give it a 5 seconds timeout.
+            // Prevent event reporting from taking too long time.
+            // Learnd from goshim's containerd/runtime/v2/shim/publisher.go
             publisher
-                .publish(Context::default(), &topic, &ns, e)
+                .publish(with_timeout(5000000000), &topic, &ns, e)
                 .await
                 .unwrap_or_else(|e| warn!("publish {} to containerd: {}", topic, e));
         }
--- a/crates/shim/src/asynchronous/publisher.rs
+++ b/crates/shim/src/asynchronous/publisher.rs
@@ -16,24 +16,42 @@
 
 use std::os::unix::io::RawFd;
 
-use async_trait::async_trait;
 use containerd_shim_protos::{
-    api::Empty,
+    api::Envelope,
     protobuf::MessageDyn,
     shim::events,
-    shim_async::{Client, Events, EventsClient},
+    shim_async::{Client, EventsClient},
     ttrpc,
-    ttrpc::{context::Context, r#async::TtrpcContext},
+    ttrpc::context::Context,
 };
+use log::debug;
+use tokio::sync::mpsc;
 
 use crate::{
-    error::Result,
+    error::{self, Result},
     util::{asyncify, connect, convert_to_any, timestamp},
 };
 
+/// The publisher reports events and uses a queue to retry the event reporting.
+/// The maximum number of attempts to report is 5 times.
+/// When the ttrpc client fails to report, it attempts to reconnect to the client and report.
+
+/// Max queue size
+const QUEUE_SIZE: i64 = 1024;
+/// Max try five times
+const MAX_REQUEUE: i64 = 5;
+
 /// Async Remote publisher connects to containerd's TTRPC endpoint to publish events from shim.
 pub struct RemotePublisher {
-    client: EventsClient,
+    pub address: String,
+    sender: mpsc::Sender<Item>,
+}
+
+#[derive(Clone, Debug)]
+pub struct Item {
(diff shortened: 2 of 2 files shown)


jokemanfire:
#312

jokemanfire:
CI test has some problem ..  Hello @mxpv

Mossaka:
need a rebase, the CI problem was resolved in #321

crates/runc-shim/src/service.rs
```diff
@@ -219,7 +219,7 @@ async fn forward(
     tokio::spawn(async move {
         while let Some((topic, e)) = rx.recv().await {
             publisher
-                .publish(Context::default(), &topic, &ns, e)
+                .publish(with_timeout(5000000000), &topic, &ns, e)
```
mxpv, line 222: Can you clarify this? Why?

crates/shim/src/asynchronous/publisher.rs
```diff
@@ -18,34 +18,110 @@ use std::os::unix::io::RawFd;
...
+use tokio::sync::{mpsc, Mutex};
 
 use crate::{
-    error::Result,
+    asynchronous::Arc,
+    error::{self, Result},
     util::{asyncify, connect, convert_to_any, timestamp},
 };
+//Mas queue size
+const QUEUE_SIZE: i64 = 1024;
```
mxpv, line 37: Please use rust's doc here.

crates/shim/src/asynchronous/publisher.rs
```diff
@@ -18,34 +18,110 @@ use std::os::unix::io::RawFd;
...
+        let receiver = self.receiver.clone();
+        let client = self.client.clone();
+        let sender = self.sender.clone();
+        let address = self.address.clone();
+        tokio::spawn(async move {
+            //only this use receiver
+            let mut recver = receiver.lock().await;
+            while let Some(item) = recver.recv().await {
+                if item.count > MAX_REQUEUE {
+                    debug!("drop event");
```
mxpv, line 85:
```suggestion
                    debug!("drop event {:?}", item);
```

crates/shim/src/asynchronous/publisher.rs
```diff
@@ -18,34 +18,110 @@ use std::os::unix::io::RawFd;
...
-        let client = Self::connect(address).await?;
+        let client = Self::connect(address.as_ref()).await?;
+        let (sender, receiver) = mpsc::channel::<Item>(QUEUE_SIZE as usize);
+        let rt = RemotePublisher {
+            address: address.as_ref().to_string(),
+            client: Arc::new(Mutex::new(EventsClient::new(client))),
+            sender,
+            receiver: Arc::new(Mutex::new(receiver)),
+        };
+        rt.process_queue().await;
```
mxpv, line 68: You can pass receiver directly to the process queue thread. This way you can remove Arc and Mutex.

```suggestion
        rt.process_queue(client).await;
```

Same applies to client. mpsc assumes one receiver, so you can use client there without locking.

crates/shim/src/asynchronous/publisher.rs
```diff
@@ -76,10 +152,22 @@ impl RemotePublisher {
...
+        let item = Item {
+            ev: envelope.clone(),
+            ctx: ctx.clone(),
+            count: 0,
+        };
         let mut req = events::ForwardRequest::new();
         req.set_envelope(envelope);
 
-        self.client.forward(ctx, &req).await?;
+        if let Err(e) = self.client.lock().await.forward(ctx, &req).await {
```
mxpv, line 163: should you enqueue here, so the background queue process it?

jokemanfire: I means that if main thread client send fail , only this situation will push it in the queue, and try to resend it. So I use mutex and arc for main thread and `process queue ` task can use the client.  If all send to the queue ,the client will not need mutex and arc . This design is for the new event will not wait for the queue to deal. This design is same with goshim/publisher.go.

crates/shim/src/asynchronous/publisher.rs
jokemanfire: Use the Arc and Mutex is that the other thread will use it.

crates/runc-shim/src/service.rs
jokemanfire: In our product environment , we've found same ttrpc request will hold on for a long time, it will cause same shim problem, learn from goshim, make it more elegent, but the five seconds also like the go shim.

mxpv: This deserves a comment

crates/shim/src/asynchronous/publisher.rs
mxpv: Right, you don't need it in the other thread. Just the process queue thread.

crates/shim/src/asynchronous/publisher.rs
```diff
@@ -76,37 +141,36 @@ impl RemotePublisher {
...
+        let item = Item {
+            ev: envelope.clone(),
+            ctx: ctx.clone(),
+            count: 0,
+        };
         let mut req = events::ForwardRequest::new();
         req.set_envelope(envelope);
 
-        self.client.forward(ctx, &req).await?;
+        //if channel is full and send fail ,release it after 3 seconds
```
mxpv, line 152: `req` seems to be no longer used here?

crates/shim/src/asynchronous/publisher.rs
mxpv: What I meant here is that comment should start with `///` or `//!` to be properly recognized by rust's doc system.

The one that you added (`#doc = "..."`) should probably just go to the file's top doc comment (`//!`) to represent module documentation.

mxpv (approved):

crates/shim/src/asynchronous/publisher.rs
```diff
@@ -76,37 +148,34 @@ impl RemotePublisher {
...
+            .send_timeout(item, tokio::time::Duration::from_secs(3))
+            .await
+            .map_err(|e| error::Error::Ttrpc(ttrpc::error::Error::Others(e.to_string())))?;
 
         Ok(())
     }
 }
 
-#[async_trait]
-impl Events for RemotePublisher {
```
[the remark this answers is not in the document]
jokemanfire: Because this trait interface is not used by async publisher , so I removed it . I use the ttrpc's forward's interface dircetly
`if let Err(e) = client.forward(new_item.ctx.clone(), &req).await {`
Is any problems caused by here?This interface should no longer be used externally.

PR #321, opened 2024-10-03
Fix compatibility with latest ttrpc and pin ttrpc version to 0.8.2

mxpv:
Looks like new version of ttrpc introduces some backward incompatible changes (https://github.com/containerd/ttrpc-rust/pull/235), this PR makes a few fixes in `shim` and `shim-protos` crates.

cc: @Mossaka 

ref: https://github.com/containerd/rust-extensions/pull/317


Mossaka (approved):

jsturtevant:
LGTM

ningmingxiao:
I don't make sure if other libs is updated, we can build the same binary(same sha256). We should add some like go.sum to make sure we can  replicate.

PR #329, opened 2024-10-24
ci:add cri-integration

ningmingxiao:


jsturtevant:
@ningmingxiao We already have an integration step in the workflow that does the setup and runs the tests. What is the purpose of adding this additional step?

ningmingxiao:
because containerd have make integration and make cri-integration 
https://github.com/containerd/containerd/blob/main/BUILDING.md
make integration  is not enough.

PR #331, opened 2024-10-28
No need to write address twice

jokemanfire:
Optimize code

### DIFF
--- a/crates/shim/src/asynchronous/mod.rs
+++ b/crates/shim/src/asynchronous/mod.rs
@@ -54,7 +54,7 @@ use crate::{
     asynchronous::{monitor::monitor_notify_by_pid, publisher::RemotePublisher},
     error::{Error, Result},
     logger, parse_sockaddr, reap, socket_address,
-    util::{asyncify, read_file_to_str, write_str_to_file},
+    util::{asyncify, read_file_to_str},
     Config, Flags, StartOpts, SOCKET_FD, TTRPC_ADDRESS,
 };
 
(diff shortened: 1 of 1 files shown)


crates/shim/src/asynchronous/mod.rs
```diff
@@ -260,8 +260,7 @@ pub async fn spawn(opts: StartOpts, grouping: &str, vars: Vec<(&str, &str)>) ->
                     return Err(e);
                 };
             }
-            if let Ok(()) = wait_socket_working(&address, 5, 200).await {
-                write_str_to_file("address", &address).await?;
+            if wait_socket_working(&address, 5, 200).await.is_ok() {
```
mxpv, line 263: Where is the first write?

jokemanfire: while the service's start_shim call spawn , if the spawn return ok  ,will write address again.

jsturtevant: @jokemanfire this isn't the same as writing it to the std_out.  This is writing it to the oci bundle path as a file.  I believe this is required so that if containerd restarts it knows how to reconnect to the shim:

https://github.com/containerd/containerd/blob/29ecab63c8c200f286a2439ea6bdddfa3d4fcaa8/cmd/containerd-shim-runc-v2/task/service.go#L102-L106

jokemanfire: Thanks @jsturtevant  , I know the socket address will be written  but I feel confuse about this [rshim](https://github.com/containerd/rust-extensions/blob/main/crates/runc-shim/src/service.rs)
`        let vars: Vec<(&str, &str)> = vec![("THP_DISABLED", thp_disabled.as_str())];

        let address = spawn(opts, &grouping, vars).await?;
        write_str_to_file("address", &address).await?;
`
also write address to address's file again .
but the func spawn return type is ok.

analytically: I'm seeing two times "false" written to my container logs?

PR #326, opened 2024-10-11
fix os_pipe doesn't work with async IO

ningmingxiao:
fix https://github.com/oconnor663/os_pipe.rs/issues/35

PR #335, opened 2024-11-05
runc: split Pipe, Io, and PipedIo to async and sync modules

Mossaka:
An attempt to resolve https://github.com/containerd/rust-extensions/pull/326#issuecomment-2458210761

Questions: should we add `cargo build -p runc` to the CI since the CI is not building runc with default target at all? 

Can you take a look at this PR? @ningmingxiao, @mxpv

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,17 +22,20 @@ jobs:
       - run: ./scripts/install-protobuf.sh
         shell: bash
 
-      - run: cargo check --examples --tests --all-targets
-      - run: cargo check --examples --tests --all-targets --all-features
-
       - run: rustup toolchain install nightly --component rustfmt
       - run: cargo +nightly fmt --all -- --check --files-with-diff
 
+      # the "runc" and "containerd-shim" crates have `sync` code that is not covered by the workspace
+      - run: cargo check -p runc --all-targets
+      - run: cargo clippy -p runc --all-targets -- -D warnings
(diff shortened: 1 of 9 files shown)


jokemanfire:
How about do not use 'os_pipe' but use 'tokio pipe'?   I think this library has more and more stable maintainers

Mossaka:
My goal is to do a refactoring so that the code can compile. I am happy to rebase on your Pr if that works

jokemanfire:
> My goal is to do a refactoring so that the code can compile. I am happy to rebase on your Pr if that works

I think I should wait for your split , because I forget the sync feature too, that pipe problem is only existing in async feature.

mxpv:
> Questions: should we add cargo build -p runc to the CI since the CI is not building runc with default target at all?

I wonder why is it not included. On Linux, the entire Cargo workspace should be included?

Mossaka:
I was AFK this entire week. Will follow up to address the comments when I get back to work.

Mossaka:
> I think I should wait for your spli

Okay, I will leave my PR to merge as it is, and the issue regarding FIFO could be left to a follow up. I don't want to block your PR

Mossaka:
> I wonder why is it not included. On Linux, the entire Cargo workspace should be included?

I think we are missing a `--workspace` flag on `ci.yml`

Mossaka:
Can we merge this PR now? @mxpv

mxpv (approved):

PR #339, opened 2024-11-13
Attempt to change mutex to rwlock

jokemanfire:
In high concurrency environments, some operations are read-only and do not require mutex write locks.So in tasks that only require reading locks, there will be no mutual blocking.

### DIFF
--- a/crates/runc-shim/src/task.rs
+++ b/crates/runc-shim/src/task.rs
@@ -40,7 +40,9 @@ use containerd_shim::{
 };
 use log::{debug, info, warn};
 use oci_spec::runtime::LinuxResources;
-use tokio::sync::{mpsc::Sender, MappedMutexGuard, Mutex, MutexGuard};
+use tokio::sync::{
+    mpsc::Sender, RwLock, RwLockMappedWriteGuard, RwLockReadGuard, RwLockWriteGuard,
+};
 
 use super::container::{Container, ContainerFactory};
 type EventSender = Sender<(String, Box<dyn MessageDyn>)>;
@@ -73,7 +75,10 @@ use crate::cgroup_memory;
 /// parameter of `Service`, and implements their own `ContainerFactory` and `Container`.
 pub struct TaskService<F, C> {
     pub factory: F,
-    pub containers: Arc<Mutex<HashMap<String, C>>>,
+    // In comparison, a Mutex does not distinguish between readers or writers that acquire the lock,
+    // therefore causing any tasks waiting for the lock to become available to yield.
+    // An RwLock will allow any number of readers to acquire the lock as long as a writer is not holding the lock.
+    pub containers: Arc<RwLock<HashMap<String, C>>>,
     pub namespace: String,
     pub exit: Arc<ExitSignal>,
     pub tx: EventSender,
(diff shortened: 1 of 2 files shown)


crates/runc-shim/src/task.rs
```diff
@@ -95,15 +95,27 @@ where
 }
 
 impl<F, C> TaskService<F, C> {
-    pub async fn get_container(&self, id: &str) -> TtrpcResult<MappedMutexGuard<'_, C>> {
-        let mut containers = self.containers.lock().await;
+    pub async fn get_container_write(&self, id: &str) -> TtrpcResult<RwLockMappedWriteGuard<'_, C>> {
```
mxpv, line 98: I'd suggest to name this `container` and `container_mut` to be closed to Rust's naming conventions.

jokemanfire: I've changed, but I don't know , if we can give a timeout in read_lock? If some write lock is hang on , but read lock can return a error?

jokemanfire:
Here is a simple way to reproduce , setting the tokio runtime to max_blocking_threads(like 10), start 10 container . And a deadlock will occur here, the crictl stats will not return and will hang on forever. So I add a timeout in read lock, then the stats will not block. I think the write method if not complete , but we should not block on the read side(read side return a error(deadlock) is accept).
friendly ping @mxpv

crates/runc-shim/src/task.rs
```diff
@@ -95,15 +100,27 @@ where
 }
 
 impl<F, C> TaskService<F, C> {
-    pub async fn get_container(&self, id: &str) -> TtrpcResult<MappedMutexGuard<'_, C>> {
-        let mut containers = self.containers.lock().await;
+    pub async fn get_container_mut(&self, id: &str) -> TtrpcResult<RwLockMappedWriteGuard<'_, C>> {
```
mxpv, line 103:
```suggestion
    pub async fn container_mut(&self, id: &str) -> TtrpcResult<RwLockMappedWriteGuard<'_, C>> {
```

jokemanfire: but the follow is 
The get naming is used only when there is a single and obvious thing that could reasonably be gotten by a getter. For example [Cell::get](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.get) accesses the content of a Cell.

For getters that do runtime validation such as bounds checking, consider adding unsafe _unchecked variants. Typically those will have the following signatures.

> fn get(&self, index: K) -> Option<&V>;
> fn get_mut(&mut self, index: K) -> Option<&mut V>;
> unsafe fn get_unchecked(&self, index: K) -> &V;
> unsafe fn get_unchecked_mut(&mut self, index: K) -> &mut V;

I have some doubts, but I will make the necessary changes first

mxpv (approved):
Thanks!

jokemanfire:
Hello  it looks here's some licenses problem  @mxpv 
```
error[rejected]: failed to satisfy license requirements
  ┌─ registry+https://github.com/rust-lang/crates.io-index#unicode-ident@1.0.14:4:36
  │
4 │ license = "(MIT OR Apache-2.0) AND Unicode-3.0"
  │            ------------------------^^^^^^^^^^^
  │            │                       │
  │            │                       rejected: not explicitly allowed
  │            license expression retrieved via Cargo.toml `license`
```

PR #342, opened 2024-11-22
The task_dir successfully cleans when the file is absent

### LINKED ISSUE #343: There is a difference between rshim and goshim when cleaning the task directory
When containerd is stopped, if the shim process has died and it cannot clean up the task's working directory(`WORKPATH=/var/lib/containerd/io.containerd.runtime.v2.task/{NS}/${CID}`).
restarting containerd will call `cleanupAfterDeadShim` to start a `shim-delete` process to clean up the corresponding `bundle` remnants. 
However, if the `$WORKPATH/runtime or `$WORKPATH/init.pid` is missing, it will report an error and return directly, resulting in cleanup failure. The error message is as follows:
```
="failed to clean up after shim disconnected" error="io.containerd.runc.v2-rs: IoError { context: \"failed to open file /run/containerd/io.containerd.runtime.v2.task/k8s.io/c120341d22ebda4fbd7306c6f0e50dc2b7a7a2f11fadf501fd83bdc2dd6ed0a4/runtime\", err: Os { code: 2, kind: NotFound, message: \"No such file or directory\" } }\n: exit status 1" id=c120341d22ebda4fbd7306c6f0e50dc2b7a7a2f11fadf501fd83bdc2dd6ed0a4 namespace=k8s.io
```
On the go-shim side, for the case of the missing init.pid file, it logs the situation and continues with the cleanup operation, ensuring successful cleanup.

https://github.com/containerd/containerd/blob/f0ebbd3c8a662656b458e7e1cca059a690d557a7/cmd/containerd-shim-runc-v2/manager/manager_linux.go#L318-L321
```
pid, err := runcC.ReadPidFile(filepath.Join(path, process.InitPidFile))
	if err != nil {
		log.G(ctx).WithError(err).Warn("failed to read init pid file")
	}
```
On the go-shim side, for the case of the missing `option.json` file,  no error is returned, the cleanup process continues.
https://github.com/containerd/containerd/blob/f6933cbf0824e6c7f9d126db9e24ab9191c341c7/cmd/containerd-shim-runc-v2/runc/container.go#L175-L181
```
func ReadOptions(path string) (*options.Options, error) {
	filePath := filepath.Join(path, optionsFilename)
	if _, err := os.Stat(filePath); err != nil {
		if os.IsNotExist(err) {
			return nil, nil
		}
		return nil, err
	}
       //..........
}
```

fengwei0328:
fix #343

### DIFF
--- a/crates/runc-shim/src/service.rs
+++ b/crates/runc-shim/src/service.rs
@@ -108,7 +108,7 @@ impl Shim for Service {
         let namespace = self.namespace.as_str();
         let bundle = current_dir().map_err(io_error!(e, "get current dir"))?;
         let opts = read_options(&bundle).await?;
-        let runtime = read_runtime(&bundle).await?;
+        let runtime = read_runtime(&bundle).await.unwrap_or_default();
 
         let runc = create_runc(
             &runtime,
@@ -117,7 +117,9 @@ impl Shim for Service {
             &opts,
             Some(Arc::new(ShimExecutor::default())),
         )?;
-        let pid = read_pid_from_file(&bundle.join(INIT_PID_FILE)).await?;
+        let pid = read_pid_from_file(&bundle.join(INIT_PID_FILE))
+            .await
+            .unwrap_or_default();
 
         runc.delete(&self.id, Some(&DeleteOpts { force: true }))
             .await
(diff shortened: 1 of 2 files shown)


mxpv (approved):

mxpv:
Could pls you rebase PR to fix CI?

fengwei0328:
> Could pls you rebase PR to fix CI?

yeah, already done

PR #346, opened 2024-11-28
Attemp to resolve deadlock in blocking fifo open

jokemanfire:
avoid fifo open deadlock
refs:https://users.rust-lang.org/t/fifos-problem-in-rust-if-it-should-be-kernels-problem/121779
If read side is  not open , the write side will be blocking forever, though the fifo file has been deleted.

### DIFF
--- a/crates/runc-shim/src/common.rs
+++ b/crates/runc-shim/src/common.rs
@@ -17,6 +17,7 @@
 use std::{
     env,
     fs::File,
+    future::Future,
     io::IoSliceMut,
     ops::Deref,
     os::{
@@ -25,6 +26,7 @@ use std::{
     },
     path::Path,
     sync::Arc,
+    time::Duration,
 };
 
 use containerd_shim::{
@@ -58,6 +60,8 @@ pub const GROUP_LABELS: [&str; 2] = [
 pub const INIT_PID_FILE: &str = "init.pid";
 pub const LOG_JSON_FILE: &str = "log.json";
 pub const FIFO_SCHEME: &str = "fifo";
+
+const TIMEOUT_DURATION: std::time::Duration = Duration::from_secs(3);
 
 #[derive(Deserialize)]
 pub struct Log {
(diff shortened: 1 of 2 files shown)


analytically:
The pipe-specific version is generally better when you know you're dealing with a Unix pipe, as it's more semantically correct and handles pipe-specific behaviors more robustly. You'd typically use this when building applications that need to interact with other processes through pipes or when dealing with stdin/stdout in an async context. Nicely done!

jokemanfire:
> The pipe-specific version is generally better when you know you're dealing with a Unix pipe, as it's more semantically correct and handles pipe-specific behaviors more robustly. You'd typically use this when building applications that need to interact with other processes through pipes or when dealing with stdin/stdout in an async context. Nicely done!

I have discovered this issue, but there is still no good way to avoid this deadlock. The submission is non blocking and cannot meet the requirements. If you have any ideas to avoid it from the kernel layer or user mode, thank you very much.

jokemanfire:
> Your `OpenOptions::new().open_receiver` change can also be made to the `copy_console` method

Thanks , but this resolution still in progress,  please continue to follow https://users.rust-lang.org/t/fifos-problem-in-rust-if-it-should-be-kernels-problem/121779

analytically:
True, this is a deeper architectural issue with containerd's FIFO handling

jokemanfire:
> True, this is a deeper architectural issue with containerd's FIFO handling

Now I want to solve the problem of concurrency caused by Fifo in Blocking mode. So if  you have any ideas, thank you very much. 
Notice  that , I found that it doesn't always cause a FIFO read or write deadlock when blocking the card owner.

analytically (approved):

crates/runc-shim/src/common.rs
```diff
@@ -248,3 +251,20 @@ pub(crate) fn xdg_runtime_dir() -> String {
     env::var("XDG_RUNTIME_DIR")
         .unwrap_or_else(|_| env::temp_dir().to_str().unwrap_or(".").to_string())
 }
+
+pub async fn handle_file_open<F, Fut>(
+    file_op: F,
+) -> core::result::Result<tokio::fs::File, tokio::io::Error>
```
mxpv, line 257:
```suggestion
) -> Result<tokio::fs::File, tokio::io::Error>
```

crates/runc-shim/src/common.rs
```diff
@@ -248,3 +251,20 @@ pub(crate) fn xdg_runtime_dir() -> String {
...
+where
+    F: FnOnce() -> Fut,
+    Fut: std::future::Future<Output = core::result::Result<tokio::fs::File, tokio::io::Error>>
```
mxpv, line 260:
```suggestion
    Fut: Future<Output = Result<tokio::fs::File, tokio::io::Error>>
```

mxpv (approved):

PR #347, opened 2024-11-29
Use tokio::io::copy_buf for spawn_copy

analytically:
Helps to avoid blocking, reduces system calls, can stop more cleanly between read/write cycles, better backpressure.


jokemanfire:
I don't know if tokio:: io::copy uses zero copy and if copy_buff will pass through the user state buffer once?

PR #349, opened 2024-12-03
fix: correct container mutex handling in TaskService

analytically:


mxpv:
Looks like this is the same as https://github.com/containerd/rust-extensions/pull/339 ?

PR #350, opened 2024-12-04
Optimize: refine the scope of the lock.

jokemanfire:
We can discard the lock before reporting the incident.According to Goshim.

### DIFF
--- a/crates/runc-shim/src/task.rs
+++ b/crates/runc-shim/src/task.rs
@@ -191,17 +191,17 @@ where
         info!("Create request for {:?}", &req);
         // Note: Get containers here is for getting the lock,
         // to make sure no other threads manipulate the containers metadata;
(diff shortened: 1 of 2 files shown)


crates/runc-shim/src/task.rs
```diff
@@ -203,7 +203,7 @@ where
         resp.pid = pid;
 
         containers.insert(id.to_string(), container);
-
+        drop(containers);
```
mxpv, line 206: I'm pretty sure most of these statements can be rewritten in the chain form like this:

```rust
        self.containers
            .write()
            .await
            .insert(id.to_string(), container);
```

so we don't have to manually call drop everywhere.

mxpv (approved):
Nice!

mxpv:
Looks like this needs to be rebased before merging.

PR #351, opened 2024-12-04
Prevent the init process from exiting and continuing with start

jokemanfire:
No need to wait for runc return an error.
provent this error 

`error="failed to exec in container: failed to start exec \"af85dfb98d747b7bc952faa4ffeffb782aee6765c0c807b5ce7a999dbb774e5e\": Others(\"Other: OCI runtime exec failed: exec failed: cannot exec in a stopped container\"): unknown"
`

### DIFF
--- a/crates/runc-shim/src/task.rs
+++ b/crates/runc-shim/src/task.rs
@@ -30,10 +30,9 @@ use containerd_shim::{
             PidsResponse, StatsRequest, StatsResponse, UpdateTaskRequest,
         },
         events::task::{TaskCreate, TaskDelete, TaskExecAdded, TaskExecStarted, TaskIO, TaskStart},
-        protobuf::MessageDyn,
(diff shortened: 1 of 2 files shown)


crates/runc-shim/src/task.rs
```diff
@@ -228,6 +228,15 @@ where
     async fn start(&self, _ctx: &TtrpcContext, req: StartRequest) -> TtrpcResult<StartResponse> {
         info!("Start request for {:?}", &req);
         let mut container = self.container_mut(req.id()).await?;
+        // Prevent the init process from exiting and continuing with start
+        // Return early to reduce the time it takes to return only when runc encounters an error
+        if container.is_init_stopped().await {
```
mxpv, line 233: Can we make this more generic?

```rust
if container.init_state().await == EnumOrUnknown::new(Status::STOPPED) { ... }
```

mxpv (approved):

PR #356, opened 2025-01-06
Remove O_NONBLOCK.

### LINKED ISSUE #353: Using rust-extensions makes stdin non-blocking?
I'm running containers and consuming stdin but I'm getting `cat: -: Resource temporarily unavailable` often, but I'm unable to consistently reproduce the problem? Using git latest on AL2023.

analytically:
O_NONBLOCK on stdin is actually causing issues when reading. When we try to read from stdin (like with `cat`), it's failing immediately with EAGAIN ("Resource temporarily unavailable") because there's no data immediately available.

For stdin, we typically want blocking behavior because:

- The reader (like `cat`) expects to be able to block waiting for input
- There's no risk of deadlock on the open operation for stdin since we're opening an already-existing FIFO

Closes #353

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,38 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: "0 0 * * *" # Every day at midnight
+  pull_request:
+    paths:
+      - '.github/workflows/stale.yml'
+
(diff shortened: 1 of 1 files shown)


analytically:
@mxpv

mxpv:
I've fixed the CI in `main`, could you pls rebase your PR?

PR #357, opened 2025-01-06
Remove O_NONBLOCK.

analytically:
@mxpv

### DIFF
--- a/crates/runc/src/asynchronous/io.rs
+++ b/crates/runc/src/asynchronous/io.rs
@@ -1,0 +1,219 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+use std::{fmt::Debug, io::Result, os::unix::io::AsRawFd, process::Stdio};
+
+use async_trait::async_trait;
+use nix::unistd::{Gid, Uid};
+use tokio::fs::OpenOptions;
+
+pub use crate::Io;
+use crate::{Command, Pipe, PipedIo};
+
+#[derive(Debug, Clone)]
+pub struct IOOption {
+    pub open_stdin: bool,
+    pub open_stdout: bool,
+    pub open_stderr: bool,
+}
+
+impl Default for IOOption {
+    fn default() -> Self {
+        Self {
+            open_stdin: true,
+            open_stdout: true,
+            open_stderr: true,
+        }
+    }
+}
+
+impl PipedIo {
+    pub fn new(uid: u32, gid: u32, opts: &IOOption) -> std::io::Result<Self> {
+        Ok(Self {
+            stdin: if opts.open_stdin {
+                Self::create_pipe(uid, gid, true)?
+            } else {
+                None
+            },
+            stdout: if opts.open_stdout {
+                Self::create_pipe(uid, gid, true)?
+            } else {
+                None
+            },
+            stderr: if opts.open_stderr {
+                Self::create_pipe(uid, gid, true)?
+            } else {
+                None
+            },
+        })
+    }
+
+    fn create_pipe(uid: u32, gid: u32, stdin: bool) -> std::io::Result<Option<Pipe>> {
+        let pipe = Pipe::new()?;
+        let uid = Some(Uid::from_raw(uid));
+        let gid = Some(Gid::from_raw(gid));
+        if stdin {
+            let rd = pipe.rd.try_clone()?;
+            nix::unistd::fchown(rd.as_raw_fd(), uid, gid)?;
+        } else {
+            let wr = pipe.wr.try_clone()?;
+            nix::unistd::fchown(wr.as_raw_fd(), uid, gid)?;
+        }
+        Ok(Some(pipe))
+    }
+}
+
+/// IO driver to direct output/error messages to /dev/null.
+///
+/// With this Io driver, all methods of [crate::Runc] can't capture the output/error messages.
+#[derive(Debug)]
+pub struct NullIo {
+    dev_null: std::sync::Mutex<Option<std::fs::File>>,
+}
+
+impl NullIo {
+    pub fn new() -> std::io::Result<Self> {
+        let f = std::fs::OpenOptions::new().read(true).open("/dev/null")?;
+        let dev_null = std::sync::Mutex::new(Some(f));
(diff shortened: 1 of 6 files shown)


mxpv:
Looks like this affects both integration and formatting checks. Latest `main` is green.

jokemanfire:
@analytically  It's may not CI’s problem you should open stdin write side to make sure read side will not block
code like this 


```
  if !init.stdio.stdin.is_empty() {
            let stdin_clone = init.stdio.stdin.clone();
            let stdin_w = init.stdin.clone();
            // Open the write side in advance to make sure read side will not block,
            // open it in another thread otherwise it will block too.
            tokio::spawn(async move {
                if let Ok(stdin_w_file) = OpenOptions::new()
                    .write(true)
                    // .custom_flags(libc::O_NONBLOCK)
                    .open(stdin_clone.as_str())
                    .await
                {
                    let mut lock_guard = stdin_w.lock().unwrap();
                    *lock_guard = Some(stdin_w_file);
                }
            });
        }
```

And because the set is sync so I have to add a timeout   ` tokio::time::sleep(Duration::from_secs(1)).await;` 
before 
```
let resp = init
            .lifecycle
            .runtime
            .create(&id, bundle, Some(&create_opts))
            .await;
```

We may should change io set to async trait  next step @mxpv

jokemanfire:
This code may you should change ,and you have to improve it 
```
    async fn do_create(&self, init: &mut InitProcess, _config: CreateConfig) -> Result<()> {
        let id = init.id.to_string();
        let stdio = &init.stdio;
        let opts = &init.lifecycle.opts;
        let bundle = &init.lifecycle.bundle;
        let pid_path = Path::new(bundle).join(INIT_PID_FILE);
        let mut create_opts = runc::options::CreateOpts::new()
            .pid_file(&pid_path)
            .no_pivot(opts.no_pivot_root)
            .no_new_keyring(opts.no_new_keyring)
            .detach(false);
        let (socket, pio) = if stdio.terminal {
            let s = ConsoleSocket::new().await?;
            create_opts.console_socket = Some(s.path.to_owned());
            (Some(s), None)
        } else {
            let pio = create_io(&id, opts.io_uid, opts.io_gid, stdio)?;
            create_opts.io = pio.io.as_ref().cloned();
            (None, Some(pio))
        };

        if !init.stdio.stdin.is_empty() {
            let stdin_clone = init.stdio.stdin.clone();
            let stdin_w = init.stdin.clone();
            // Open the write side in advance to make sure read side will not block,
            // open it in another thread otherwise it will block too.
            tokio::spawn(async move {
                if let Ok(stdin_w_file) = OpenOptions::new()
                    .write(true)
                    // .custom_flags(libc::O_NONBLOCK)
                    .open(stdin_clone.as_str())
                    .await
                {
                    let mut lock_guard = stdin_w.lock().unwrap();
                    *lock_guard = Some(stdin_w_file);
                }
            });
        }
        tokio::time::sleep(Duration::from_secs(1)).await;
        let resp = init
            .lifecycle
            .runtime
            .create(&id, bundle, Some(&create_opts))
            .await;
        if let Err(e) = resp {
            if let Some(s) = socket {
                s.clean().await;
            }
            return Err(runtime_error(bundle, e, "OCI runtime create failed").await);
        }
        copy_io_or_console(init, socket, pio, init.lifecycle.exit_signal.clone()).await?;
        let pid = read_file_to_str(pid_path).await?.parse::<i32>()?;
        init.pid = pid;
        Ok(())
    }
```

analytically:
So in my setup, the parent process (concourse-worker) has already opened the write end of the FIFO. Concourse resource containers expect to be able to block on stdin reads when waiting for input.

analytically:
Just removing NONBLOCK makes rust-extensions work flawlessly with Concourse CI (https://concourse-ci.org/). I overwrite the bundled containerd-shim-runc-v2-rs and am able to run > 500 containers on a single host, with the Go version I cannot go more than 180.

jokemanfire:
> So in my setup, the parent process (concourse-worker) has already opened the write end of the FIFO. Concourse resource containers expect to be able to block on stdin reads when waiting for input.

In the contaienrd's CI the write side may not open, I've pasted your change and run it in CI , Open write side in shim side can pass this CI.

analytically:
In production Concourse usage, the write side would be opened by shim and the read would work properly. But in CI where there's no shim/write side, the NONBLOCK flag allows the tests to proceed without hanging.

analytically:
```
// Add NONBLOCK only in CI environment
if std::env::var("CI").is_ok() {
  options.custom_flags(libc::O_NONBLOCK);
}
```

jokemanfire:
> In production Concourse usage, the write side would be opened by shim and the read would work properly. But in CI where there's no shim/write side, the NONBLOCK flag allows the tests to proceed without hanging.

if it means you want to do a sync in stdin write and read. Blocking in stdin's read may not greate.
How about use pipe in middle,  may it can resolve this problem greater?

jokemanfire:
> Just removing NONBLOCK makes rust-extensions work flawlessly with Concourse CI (https://concourse-ci.org/). I overwrite the bundled containerd-shim-runc-v2-rs and am able to run > 500 containers on a single host, with the Go version I cannot go more than 180.



> Maybe make it configurable?
> 
> ```
> #[derive(Debug, Clone)]
> pub struct IOOption {
>     pub open_stdin: bool,
>     pub open_stdout: bool,
>     pub open_stderr: bool,
>     pub stdin_nonblock: bool,  // New configuration option
> }
> 
> impl Default for IOOption {
>     fn default() -> Self {
>         Self {
>             open_stdin: true,
>             open_stdout: true,
>             open_stderr: true,
>             stdin_nonblock: false,  // Default to blocking behavior
>         }
>     }
> }
> ```
> 
> Then in FIFO:
> 
> ```
> impl FIFO {
>     fn set(&self, cmd: &mut Command, opts: &IOOption) -> Result<()> {
>         if let Some(path) = self.stdin.as_ref() {
>             let mut options = OpenOptions::new().read(true);
>             if opts.stdin_nonblock {
>                 options.custom_flags(libc::O_NONBLOCK);
>             }
>             let stdin = options.open(path)?;
>             cmd.stdin(stdin);
>         }
>         // ...
>     }
> }
> ```

I think that's will be great , and read the  `env`  while set the IOOption. How about change the env's name "CI" to other name which can be more easily understood?

analytically:
The env name "CI" is standard for a lot of CI tools, see https://github.com/actions/runner/blob/fde5227fbfe9c61b7861cc959ebbbba62af4754b/src/Runner.Sdk/ProcessInvoker.cs#L284

jokemanfire:
> The env name "CI" is standard for a lot of CI tools, see https://github.com/actions/runner/blob/fde5227fbfe9c61b7861cc959ebbbba62af4754b/src/Runner.Sdk/ProcessInvoker.cs#L284

That's well, thanks.

jokemanfire:
> The env name "CI" is standard for a lot of CI tools, see https://github.com/actions/runner/blob/fde5227fbfe9c61b7861cc959ebbbba62af4754b/src/Runner.Sdk/ProcessInvoker.cs#L284

I've changed the IO set for Blocking IO's Time consuming issue, may need  a rebase.

PR #359, opened 2025-01-08
OpenOptions in blocking mod is a time-consuming operation.

jokemanfire:
Prevent it from blocking a tokio thread. Change sync to async.
And fix an error in Pipe struct.

### DIFF
--- a/crates/runc/src/asynchronous/io.rs
+++ b/crates/runc/src/asynchronous/io.rs
@@ -1,0 +1,219 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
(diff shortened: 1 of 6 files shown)


crates/runc/src/asynchronous/io.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    fn create_pipe(
+        uid: u32,
+        gid: u32,
+        enabled: bool,
+        stdin: bool,
+    ) -> std::io::Result<Option<Pipe>> {
+        if !enabled {
+            return Ok(None);
```
mxpv, line 59: I don't think we need `enabled: bool` param here. It should be up to caller.
You could do this instead:

```rust
Ok(Self {
    stdin: if opts.open_stdin { Self::create_pipe(uid, gid, true)?.into() } else { None },
    ...
})
```

mxpv:
Overall looks good! I have one minor comment.

PR #366, opened 2025-01-20
While start the shim fail, ensure the socket file has been remove

jokemanfire:
Discovered socket file leak when set_cgroup_ond_oom_store execution failed


jokemanfire:
Some fail I have found , while stdin can't parse 
"FATA[0001] run pod sandbox: rpc error: code = Unknown desc = failed to start sandbox "b09ff5e143f7f482d71737bd637151cec607d1908c88286552ab740386a7e928": failed to create containerd task: failed to start shim: start failed: io.containerd.runc.v2-rs: Other("add task to cgroup: unable to write to a control group file /sys/fs/cgroup/blkio/BinaryName = ''\nCriuImagePath = ''\nCriuPath = ''\nCriuWorkPath = ''\nIoGid = 0\nIoUid = 0\nNoNewKeyring = false\nNoPivotRoot = false\nRoot = ''\nShimCgroup = ''\nSystemdCgroup = false\n/cgroup.procs, value [CREATE FILE] caused by: Os { code: 2, kind: NotFound, message: \"No such file or directory\" }"

crates/shim/src/asynchronous/mod.rs
```diff
@@ -344,10 +344,18 @@ pub async fn spawn(opts: StartOpts, grouping: &str, vars: Vec<(&str, &str)>) ->
         command.arg("-debug");
     }
     command.envs(vars);
+    let result = command.spawn().map_err(io_error!(e, "spawn shim"));
+    if let Err(e) = result {
+        remove_socket(&address).await?;
+        return Err(e);
+    }
+    let _child = result?;
```
mxpv, line 352: You don't need to unpack the result twice:

```rust
let child = match result {
   Ok(child) => child,
   Err(e) => {
     remove_socket(&address).await?;
     return Err(e);
   }
```

jokemanfire: That's well

jokemanfire:
@mxpv hi, could we fix this bug now?

PR #372, opened 2025-01-29
Add example using the transfer service

estesp:
Pulls an image to the containerd content store

### DIFF
--- a/crates/client/examples/container_pull.rs
+++ b/crates/client/examples/container_pull.rs
@@ -1,0 +1,91 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+use std::env::consts;
+
+use client::{
+    services::v1::{transfer_client::TransferClient, TransferOptions, TransferRequest},
+    to_any,
+    types::{
+        transfer::{ImageStore, OciRegistry, UnpackConfiguration},
+        Platform,
+    },
+    with_namespace,
+};
+use containerd_client as client;
+use tonic::Request;
+
+const IMAGE: &str = "docker.io/library/alpine:latest";
+const NAMESPACE: &str = "default";
+
+/// Make sure you run containerd before running this example.
+/// NOTE: to run this example, you must prepare a rootfs.
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let arch = match consts::ARCH {
+        "x86_64" => "amd64",
+        "aarch64" => "arm64",
+        _ => consts::ARCH,
+    };
+
+    let channel = client::connect("/run/containerd/containerd.sock")
+        .await
+        .expect("Connect Failed");
+    let mut client = TransferClient::new(channel.clone());
+
+    // Create the source (OCIRegistry)
+    let source = OciRegistry {
+        reference: IMAGE.to_string(),
+        resolver: Default::default(),
+    };
+
+    let platform = Platform {
+        os: "linux".to_string(),
+        architecture: arch.to_string(),
+        variant: "".to_string(),
+        os_version: "".to_string(),
+    };
+
+    // Create the destination (ImageStore)
+    let destination = ImageStore {
+        name: IMAGE.to_string(),
+        platforms: vec![platform.clone()],
+        unpacks: vec![UnpackConfiguration {
+            platform: Some(platform),
+            ..Default::default()
+        }],
+        ..Default::default()
+    };
+
+    let anys = to_any(&source);
+    let anyd = to_any(&destination);
+
+    println!("Pulling image for linux/{} from source: {:?}", arch, source);
+
+    // Create the transfer request
+    let request = TransferRequest {
+        source: Some(anys),
+        destination: Some(anyd),
+        options: Some(TransferOptions {
+            ..Default::default()
+        }),
+    };
+    // Execute the transfer (pull)
+    client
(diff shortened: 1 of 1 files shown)


mxpv:
You can run `cargo fmt` locally like this:

```bash
❯ cargo +nightly fmt --all -- --check
```

image

estesp:
> ```shell
> cargo +nightly fmt --all -- --check
> ```

Well, I tried that command and it didn't work with a strange error about lack of toolchain directive support in cargo, hence my weird invocation via rustup 😂  Turns out I had a weird path problem with my install, so now `cargo +nightly` works properly. Thanks!

mxpv (approved):

aFuerst:
@estesp can auth credentials to a private repository be passed with the OciRegistry.resolver `headers` hashmap?
I'm trying to replicate `ctr images pull --user name:pass`

Or a pointer to the containerd could would be helpful

estesp:
I haven't taken the time to implement it yet, but this feature depends on using a gRPC stream to handle a synchronous callback from containerd's transfer service to ask for credentials.

You pass the stream to the transfer service through the [OCI registry type `auth_stream` entry in the `RegistryResolver`](https://github.com/containerd/rust-extensions/blob/main/crates/client/vendor/github.com/containerd/containerd/vendor/github.com/containerd/containerd/api/types/transfer/registry.proto#L33). Note that the general streaming API is defined [here](https://github.com/containerd/rust-extensions/blob/main/crates/client/vendor/github.com/containerd/containerd/vendor/github.com/containerd/containerd/api/services/streaming/v1/streaming.proto).

The transfer service sends an `AuthRequest` type on that stream and you would have to respond with an `AuthResponse` for that host. The headers entry is most likely just to support extra header content you wish to pass for that registry? More details on the auth flow in the transfer service doc: https://github.com/containerd/containerd/blob/main/docs/transfer.md#credentials

aFuerst:
Thanks for the pointers! I'll see if I can get it working, looks straightforward

gopilola:
> Thanks for the pointers! I'll see if I can get it working, looks straightforward

@aFuerst, were you able to implement auth credentials to a private repository? Could you share some code samples to help me get started on it.

aFuerst:
@gopilola I was never able to get it working. Here is my WIP attempt at it. A stream I would open would keep being closed before I could use it. Either auth or simpler status checking.
https://github.com/aFuerst/containerd-rust-image-pull

gopilola:
> @gopilola I was never able to get it working. Here is my WIP attempt at it. A stream I would open would keep being closed before I could use it. Either auth or simpler status checking. https://github.com/aFuerst/containerd-rust-image-pull

thanks for the response, @aFuerst. same with me!! i used your example and tried the progress stream too and that also did not work. Not sure if it related to https://github.com/containerd/containerd/pull/7886 where we need to use a lease to make sure it is not deleted before it is used. But to use it, we need to set the lease id as GRPC header and there is no provision to do that with rust extension.

```
use containerd_client::tonic::Request;
use containerd_client::services::v1::streaming_client::StreamingClient;
use containerd_client::services::v1::transfer_client::TransferClient;
use containerd_client::services::v1::TransferRequest;
use containerd_client::types::Platform;
use containerd_client::types::transfer::{ImageStore, OciRegistry, RegistryResolver, UnpackConfiguration};
use containerd_client::with_namespace;
use uuid::Uuid;

const CONTAINERD_SOCK: &str = "/run/containerd/containerd.sock";
// const IMAGE_NAME: &str = "registry.docker.iu.edu/iluvatar-faas/hello-iluvatar-action:latest";
const IMAGE_NAME: &str = "https://<container-registry>/python-fastapi-app:latest";
const NAMESPACE: &str = "default";

#[tokio::main(flavor = "multi_thread", worker_threads = 10)]
async fn main() {
    let channel = containerd_client::connect(CONTAINERD_SOCK).await.unwrap();

    let stream_uuid = Uuid::new_v4().to_string();
    let mut stream_client = StreamingClient::new(channel.clone());
    let req = containerd_client::services::v1::StreamInit {
        id: stream_uuid.clone()
    };
    let req = containerd_client::to_any(&req);
    // let req = with_namespace!(req, NAMESPACE);
    let mut stream = match stream_client.stream(tokio_stream::iter([req])).await {
        Ok(s) => s.into_inner(),
        Err(e) => {
            println!("stream init failed {:?}", e) ;
            return;
        }
    };
    println!("checking stream 1!");
    match stream.message().await {
        Err(e) => {
            println!("rcv stream init failed {:?}", e);
            return;
        },
        Ok(None) => println!("init stream closed?"),
        Ok(Some(val)) => println!("init stream value {:?}", val)
    };
    let trans_options = Some(containerd_client::services::v1::TransferOptions { progress_stream:stream_uuid.clone() });
    let resolver = Some(RegistryResolver {
        // auth_stream: stream_uuid.clone(),
        ..Default::default()
    });

    let source = OciRegistry {
        reference: IMAGE_NAME.to_string(),
        resolver,
    };
    let arch: String = match std::env::consts::ARCH {
        "x86_64" => "amd64",
        "aarch64" => "arm64",
        _ => std::env::consts::ARCH,
    }.to_string();
    let platform = Platform {
        os: "linux".to_string(),
        architecture: arch,
        ..Default::default()
    };

    let destination = ImageStore {
        name: IMAGE_NAME.to_string(),
        platforms: vec![platform.clone()],
        unpacks: vec![UnpackConfiguration {
            platform: Some(platform),
            snapshotter: "overlayfs".to_string(),
        }],
        ..Default::default()
    };

    let anys: prost_types::Any = containerd_client::to_any(&source);
    let anyd: prost_types::Any = containerd_client::to_any(&destination);
    let request = TransferRequest {
        source: Some(anys),
        destination: Some(anyd),
        options: trans_options,
    };

    println!("sending pull request");
    let j = tokio::spawn(  async move{
        let mut client = TransferClient::new(channel.clone());
        client.transfer(with_namespace!(request, NAMESPACE)).await
    });
    println!("checking stream 3!");
    match stream.message().await {
         Err(e) => println!("rcv stream init failed {:?}", e),
         Ok(None) => println!("init stream closed?"),
         Ok(Some(val)) => println!("init stream value {:?}", val)
    };
    println!("checking pull result");
    match j.await.unwrap()  {
        Ok(_) => println!("pull success!"),
        Err(e) => println!("Error pulling image {:?}", e),
    }
}
```

PR #374, opened 2025-02-08
add proto3 flag to fix cargo doc generation

kykosic:
I noticed cargo docs are failing for this crate because the protoc version in the build environment doesn't by default support optional for proto3. Adding the `--experimental_allow_proto3_optional` flag should fix this.


mxpv (approved):
Thanks!

PR #375, opened 2025-02-11
Change CI's rustup version to use rust-toolchain.toml 's version

jokemanfire:
and fix test's fail. Ensure that the CI is consistent with the actual version of the rustup toolchain built.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,12 +18,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-
+      # Use the rust-toolchain-file 
+      - uses: dsherret/rust-toolchain-file@v1
       - run: ./scripts/install-protobuf.sh
         shell: bash
 
-      - run: rustup toolchain install nightly --component rustfmt
-      - run: cargo +nightly fmt --all -- --check
+      - run: cargo fmt --all -- --check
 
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -21,9 +21,9 @@ jobs:
 
       - run: ./scripts/install-protobuf.sh
         shell: bash
-
-      - run: rustup toolchain install nightly --component rustfmt
-      - run: cargo +nightly fmt --all -- --check
+      # Remove the manual rustup nightly installation step
+      # - run: rustup toolchain install nightly --component rustfmt
```
mxpv, line 25: Just remove these lines? We have git to take care of history.

jokemanfire: I do not sure , this change is ok?

jokemanfire: ref: https://github.com/dsherret/rust-toolchain-file

mxpv:
Actually, can you add some clarity to this?
Rustup expected to pickup `rust-toolchain.toml` [automatically](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), right?

jokemanfire:
image
In CI use nighlty version check will not same with rust toolchain's version ,so may pass in local sometime ,but fail in CI. I think check should use same rustup toolchain version .

PR #380, opened 2025-02-28
make the async feature a no-op on non-unix platforms

### LINKED ISSUE #379: `containerd-shim` fails to build on Windows with the `async` feature enabled.
The async feature in Windows should be a no-op.
More generally, features should be additive.
This is not the case for the async feature, but it could be a no-op in unsupported platforms.

jprendes:
Fixes #379

### DIFF
--- a/crates/shim/src/lib.rs
+++ b/crates/shim/src/lib.rs
@@ -77,7 +77,7 @@ pub mod api {
 macro_rules! cfg_not_async {
     ($($item:item)*) => {
         $(
-            #[cfg(not(feature = "async"))]
+            #[cfg(not(all(feature = "async", target_family="unix")))]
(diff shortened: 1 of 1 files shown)


Mossaka (approved):

mxpv:
Can you add some clarity around this?

jprendes:
Depending on rust-extensions while trying to maintain support for windows and for async (whenever possible) is challenging with the current setup, as containerd-shim fails to build on windows when the async feature is enabled.
Making the feature a no-op on windows would work around this issue, allowing the use of async on unix platforms, and sync on windows.

That being said, I am currently collaborating with ttrpc-rust to add [async support for windows there](https://github.com/containerd/ttrpc-rust/pull/290).
Once that's merged I would like to give it a go to adding async support for windows here.
I think overall that would be better than what's proposed in this PR.

PR #382, opened 2025-03-12
Consolidate the Windows NamedPipeLogger into the FifoLogger

jprendes:
There's a fair amount of duplicated code between both loggers.
Given that we can treat the named pipe as a `std::fs::File`, we can consolidate the functionality of both.

This PR:
* removes the `sys` module with the `NamedPipeLogger`
* brings the `NamedPipeLogger` functionality as a new `with_named_pipe` method in the `FifoLogger`.
* moves all the `NamedPipeLogger` tests to the `FifoLogger` module
* **breaking:** changes the signature of the `FifoLogger::new` so that it can construct the windows logger.


mxpv (approved):

jprendes:
Is this ready to merge?
Otherwise, how can I help? :-)

mxpv:
@jprendes the PR looks good. It looks like the CI is broken because of some other issue, I'll try to have a look soon.

PR #383, opened 2025-03-19
Use same activation strategy in all platforms

jprendes:
Currently the sync code uses two different activation strategies:
* **In Unix:** The server listener is bound in the parent process, and passes the FD to the child processes.
* **In Windows:** The server listener is directly bound in the child process, and a pipe is used for the parent to wait until the binding is completed.

This PR unified the activation strategy on all platforms, following the strategy used for Windows.
Additionally, it avoids computing the listener address twice by reusing the `-socket` flag when spawning the child process.

This PR also applies the same activation strategy to the async code.

The main motivation for this PR is to eventually support Windows on the async feature.

### DIFF
--- a/crates/shim/Cargo.toml
+++ b/crates/shim/Cargo.toml
@@ -72,9 +72,6 @@ tokio = { workspace = true, features = ["full"], optional = true }
 [target.'cfg(target_os = "linux")'.dependencies]
 cgroups-rs.workspace = true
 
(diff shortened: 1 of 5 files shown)


jprendes:
@mxpv , PTAL

Mossaka (approved):

mxpv (approved):

PR #384, opened 2025-03-19
Unblock CI

jprendes:
This PR aims to unblock CI, which is currently failing.
This bumps the deny.toml file to the new format, and adds `proc-macro-error` and `protobuf` advisories to the ignore list.
* `proc-macro-error` is marked as unmaintained, and is not a direct dependencies of `rust-extensions`. It should be fixed upstream.
* `protobuf` is a direct dependency of `rust-extensions`, but bumping the dependency here does not fix the issue, as `ttrcp-codegen` still depends on an older version of `protobuf`.

In my opinion, in the case of `protobuf`, this should be considered a temporary solution, and we should work with upstream to fix the underlying issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: EmbarkStudios/cargo-deny-action@v1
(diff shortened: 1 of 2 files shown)


jprendes:
@mxpv PTAL :-)

Mossaka (approved):

mxpv (approved):

PR #387, opened 2025-04-08
Remove dependency on signal-hook-tokio

jprendes:
The `signal-hook-tokio` package is now integrated into tokio in the `signals` feature.
Moreover, the `signal-hook-tokio` crates is unix only (see https://github.com/vorner/signal-hook/issues/100#issuecomment-804334820), while the tokio impl also supports windows.

This PR is part of the effort to add Windows support for the async feature.

### DIFF
--- a/crates/shim/src/asynchronous/mod.rs
+++ b/crates/shim/src/asynchronous/mod.rs
@@ -15,7 +15,6 @@
 */
 
 use std::{
-    convert::TryFrom,
     env,
     io::Read,
     os::unix::{fs::FileTypeExt, net::UnixListener},
@@ -25,6 +24,7 @@ use std::{
         atomic::{AtomicBool, Ordering},
         Arc,
     },
+    task::{ready, Poll},
 };
 
 use async_trait::async_trait;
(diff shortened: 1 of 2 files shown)


jprendes:
@mxpv @jsturtevant PTAL, thanks!

crates/shim/src/asynchronous/mod.rs
```diff
@@ -388,13 +386,73 @@ fn signal_server_started() {
...
+fn signal_stream(kind: i32) -> std::io::Result<BoxStream<'static, i32>> {
+    use tokio::signal::unix::{signal, SignalKind};
+    let kind = SignalKind::from_raw(kind);
+    signal(kind).map(|mut sig| {
+        use std::task::{ready, Poll};
+
+        use futures::stream::poll_fn;
+        poll_fn(move |cx| -> Poll<Option<i32>> {
+            ready!(sig.poll_recv(cx));
+            Poll::Ready(Some(kind.as_raw_value()))
```
mxpv, line 399: I'm not sure I understand this part. Why do we poll manually here?

jprendes: An alternative would be to use `tokio-stream`'s wrappers for Signals and CtrlC: https://docs.rs/tokio-stream/latest/tokio_stream/wrappers/struct.SignalStream.html

I don't think it's worth adding a new dependency when `poll_fn` works fine, but it would be a clearer without having to add a comment.

jprendes:
Looks like there are some unrelated issues with the Ubuntu 20.04 runners.
It fails even before starting the workflow.

> This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101
>
> ---
> GitHub Actions has encountered an internal error when running your job.

Is there any specific reason to test on Ubuntu 20.04?

PR #388, opened 2025-04-08
check Cargo.lock into the version control system to prevent errors caused by yanked dependence

zhaodiaoer:
fix errors in testing code like :
```
error[E0277]: the trait bound `Box<FakeServer>: containerd_shim_protos::shim_async::Task` is not satisfied
  --> crates/shim-protos/examples/ttrpc-server-async.rs:63:32
   |
63 |     let tservice = create_task(Arc::new(Box::new(FakeServer::new())));
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `containerd_shim_protos::shim_async::Task` is not implemented for `Box<FakeServer>`
   |
   = help: the trait `containerd_shim_protos::shim_async::Task` is implemented for `FakeServer`
   = note: required for the cast from `Arc<Box<FakeServer>>` to `Arc<(dyn containerd_shim_protos::shim_async::Task + Send + Sync + 'static)>`
```


jprendes:
I don't see this error on my dev environment.

However, the underlying issue is caused by this change
https://github.com/containerd/ttrpc-rust/pull/235

That breaking change was "briefly" (for 3 months) released as 0.6.3 without bumping the major version, but that version is now yanked (as of mid January 2025), and re-released as 0.7.0.
Which version of ttrpc-compiler are you using?
What's the output of `cargo tree -i -p ttrpc`?
If you see 0.6.3, please try updating your `Cargo.lock` file and try again.

When we bump the ttrpc dependencies, then we would need to fix this breaking change as well here.

zhaodiaoer:
> I don't see this error on my dev environment.
> 
> However, the underlying issue is caused by this change [containerd/ttrpc-rust#235](https://github.com/containerd/ttrpc-rust/pull/235)
> 
> That breaking change was "briefly" (for 3 months) released as `0.6.3` without bumping the major version, but that version is now yanked (as of mid January 2025), and re-released as `0.7.0`. See [the versions page in crates.io](https://crates.io/crates/ttrpc-compiler/versions). Which version of `ttrpc-compiler` are you using? What's the output of `cargo tree -i -p ttrpc`? If you see `0.6.3`, please try updating your `Cargo.lock` file and try again.
> 
> When we bump the ttrpc dependencies, then we would need to fix this breaking change as well here.

Thanks for your context, you are right, the version of `ttrpc-compiler` in my `Cargo.lock` is precisely `0.6.3`, when i have
downgraded it to `0.6.2`, all errors have gone:

```
$ cargo tree -i -p ttrpc-compiler

ttrpc-compiler v0.6.3
└── ttrpc-codegen v0.4.2
    [build-dependencies]
    └── containerd-shim-protos v0.8.0 (/home/bytedance/code/opensource/zhaodiaoer/rust-extensions/crates/shim-protos)
        └── containerd-shim v0.8.0 (/home/bytedance/code/opensource/zhaodiaoer/rust-extensions/crates/shim)
            └── containerd-runc-shim v0.1.1 (/home/bytedance/code/opensource/zhaodiaoer/rust-extensions/crates/runc-shim)
```

But now i'm staring to be curious about another thing: should we let `Cargo.lock` out from `.gitignore`? Otherwise, we often
need to handle similar matters manually.

jprendes:
> should we let `Cargo.lock` out from `.gitignore`?

Since containerd-runc-shim is a binary crate, I think you are right and we should track `Cargo.lock`.
See [the official advice](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html).

zhaodiaoer:
@mxpv @jsturtevant PTAL, thanks!

jokemanfire:
https://github.com/containerd/rust-extensions/pull/375/files
We could use arc::from instead of arc::new to be compatible with such modifications.

zhaodiaoer:
> https://github.com/containerd/rust-extensions/pull/375/files We could use arc::from instead of arc::new to be compatible with such modifications.

Useful information, thanks!

PR #393, opened 2025-04-21
fix(update): avoid update resource while the process is zombie

jokemanfire:
add a check in zombie init process.

If update task between the wait_pid call and init exit. The bug will occur,  the cgroup maybe update the root's cgroup path.
Because the cgroup path has been recycled by system ,but the wait_pid haven't recycled the resource of the process. The update task is processing. 

A easy way to reproduce this bug. Occasionally triggered in high concurrency scenarios.
1) start a shim ,like cri
crictl run --no-pull container-config.json pod.json 
2) gdb -p {target shim pod}
(gdb) handle SIGCHLD ignore
(gdb) continue
3) kill the init process

the init process will become zombie ,  you can send update task to the shim ,but the cgroup path is not correct.

### DIFF
--- a/crates/runc-shim/src/runc.rs
+++ b/crates/runc-shim/src/runc.rs
@@ -65,6 +65,19 @@ use crate::{
     io::Stdio,
 };
 
+/// check the process is zombie
+#[cfg(target_os = "linux")]
+fn is_zombie_process(pid: i32) -> bool {
+    if let Ok(status) = std::fs::read_to_string(format!("/proc/{}/status", pid)) {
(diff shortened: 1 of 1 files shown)


jokemanfire:
@mxpv  @Mossaka  Project CI may have some problems , could fix it ?

jokemanfire:
@mxpv  fixed

mxpv (approved):

jokemanfire:
@mxpv  it looks like IT fail ,I checked the logs, but there's no clue. Can you take a look？

PR #420, opened 2025-07-21
ci: Test with gnu and musl targets

fidencio:
Then we can ensure we don't end up with a similar issue that happend with containerd-shim-protos not being buildable with musl on 0.9.0.


fidencio:
Here's an example of how this would fail: https://github.com/containerd/rust-extensions/actions/runs/16959634347/job/48069043211?pr=420

fidencio:
@Mossaka, sorry for almost a month to get to this one, but I finally made some time.  It's working as expected, as shown by the comment above (where I ran the CI after reverting d8f28555b5a69dc542a9206c3cfa0e252e116fa9).

PR #421, opened 2025-07-30
shim/loop-dev: fix setup_loop_dev returned error

zzzzzzzzzy9:
'libc::ioctl' should use the pointer of struct info. 
I don't know why this test case didn't get an error when I submitted the code before.


zzzzzzzzzy9:
@mxpv Done.

PR #440, opened 2025-10-09
fix(pty) : Use more safe package rustix to resize pty

jokemanfire:
avoid using the nix's unsafe ioctl call


jokemanfire:
> @jokemanfire yeah, absolutely. Could you slack me your email and first+last name. I'll open a PR to update maintainers file.

I sent you an email, if you haven't received it, maybe check the trash ? :)

PR #465, opened 2026-04-30
update sandbox api proto

Apokleos:
Update sandbox API protocols based on the latest containerd


tzneal:
Happened to see this after I had pushed https://github.com/containerd/rust-extensions/pull/466 . That CR includes this change as well since it updates all protos to the v2.3.0 ones.

mxpv:
Looks like this already covered.
I'm going to close this one.
Feel free to reopen if I missed something.

PR #482, opened 2026-07-14
Replace string replace workaround with extern_path

mxpv:


crates/client/build.rs
```diff
@@ -44,73 +59,39 @@ const PROTO_FILES: &[&str] = &[
...
-    "containerd.services.containers.v1",
-    "containerd.services.content.v1",
-    "containerd.services.events.v1",
 ];
 
 fn main() {
-    let mut config = tonic_prost_build::Config::new();
-    config.protoc_arg("--experimental_allow_proto3_optional");
-    config.enable_type_names();
+    let includes = &["vendor/github.com/containerd/containerd/api/", "vendor/"];
```
jokemanfire, line 65: about this flag , if we should remove it '--experimental_allow_proto3_optional'?
The recent [Protobuf 3.15 release](https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0) implies this flag is no longer required.
https://github.com/grpc/grpc-rust/pull/577

jokemanfire (approved):

